PR #1, opened 2018-03-12
Import tiff code from image crate

Robzz:
I managed to get the tiff code from the image crate to build on its own. I basically copied the tiff module from the image crate, then had to copy some more boilerplate (error/result types, etc...) to get it to actually compile. Finally, i fixed the warnings clippy found. There are a few left:

* 3 instances of ```warning: you have declared `#[inline(always)]` on `byte_order`. This is usually a bad idea``` in `stream.rs`. I would tend to agree with clippy on this one, but I'll let you decide.
* The second puzzles me more, I'm thinking of a false positive:

```
warning: redundant field names in struct initialization
   --> src/decoder/stream.rs:124:52
    |
124 |                 try!(reader.read_exact(&mut buffer[start..]));
    |                                                    ^^^^^ help: replace it with: `start`
    |
    = note: #[warn(redundant_field_names)] on by default
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.187/index.html#redundant_field_names
```

The rest compiles without warnings from rustc.

I haven't tried to get the image crate to use this crate yet but I'm opening this already so you can take a look at it early if you'd like. In particular, I don't know who wrote the original code, who should I name as the author in the cargo manifest ?


bvssvni:
Copy all the authors, just in case.

Robzz:
I finally had some time to finish this, I also opened a PR on the image crate to use this once this gets merged: https://github.com/PistonDevelopers/image/pull/783

Since last time, I added some super simple integration style tests using images from the libtiff test suite, so I added the libtiff copyright in the appropriate folder. I could remove these or use another tiff image suite if you'd like, but I didn't find many. I kept the pxm images around in case someone ever wants to improve the tests and actually compare the decoder output to the pxm images.

bvssvni:
Nice! Merging.

PR #9, opened 2018-08-17
Add support for ASCII tags

jimrybarski:
Certain tags contain ASCII strings, which are currently not supported. If the string contains a null byte, everything after it is truncated. There are some tags in the test images where several bytes exist downstream of the first null byte - it seems to be garbage but I could be misinterpreting the situation.


bvssvni:
Merging.

PR #10, opened 2018-08-27
Bug fix in `Decoder::expand_strip` + greatly improved performance for 16 bit images

zroug:
This pull request makes three changes:

 - It fixes an error where `Decoder::expand_strip` might not fill the whole buffer.
 - It greatly improves the performance for 16 bit images. (Reading many large tifs now takes minutes instead of hours.)
 - It removes the unsafe blocks in `Decoder::read_image` since I couldn't find any evidence that `ReadBytesExt::read_u16_into` guarantees that it doesn't read from the buffer. There isn't any significant performance impact due to zero initialization in this case.

Note that I think that there still are bugs in  `Decoder::read_image` when the image contains inconsistent information about the different sizes.


bvssvni:
Merging.

PR #11, opened 2018-09-13
Use typed errors instead of strings

fschutt:
Using strings as error types sidesteps everything that is great about Rusts error handling and is generally not very performance-friendly. Strings cannot be matched / destructured in case someone wants to catch a specific error case and it's generally bad practice to use strings as errors. 

This PR simply migrates from `String` to properly typed errors with proper `Display` impls.

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -4,7 +4,7 @@ use std::io::{self, Read, Seek};
 use std::collections::{HashMap};
 
 use super::stream::{ByteOrder, SmartReader, EndianReader};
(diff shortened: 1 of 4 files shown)


fschutt:
@bvssvni Is this crate still actively maintained? Just asking.

bvssvni:
@fschutt Yes, just busy. Do you have access to merge PRs?

fschutt:
No, I don't have merge access (idk if I should be trusted with this, I'd prefer to just be a normal contributor) - I just thought you could merge it or give it a review since you were the last contributor according to the commits.

PR #22, opened 2019-01-29
Fix dependecy versions

### LINKED ISSUE #21: Dependency versions are overly restrictive
Using a `*` forces downstream users to be unable to use newer versions of crates. It is usually suggested to avoid a version operator at all. Currently, this crate has:

https://github.com/PistonDevelopers/image-tiff/blob/4e3b36598d97fe0096f4696495f3f2af8f4f0148/Cargo.toml#L18-L21

Instead, I would expect this to be written as:

```toml
byteorder = "1.2"
lzw = "0.10"
num-derive = "0.2"
num-traits = "0.2"
```

This would allow for byteorder 1.2.x and 1.3.x, a semver-compatible version. It would **not** allow num-derive 0.3 (which is not semver-compatible with 0.2).

See [the Cargo doc for details on the operators](https://doc.rust-lang.org/stable/cargo/reference/specifying-dependencies.html).

The playground uses the newest version of crates. The image crate is popular enough to be included on the playground and it depends on tiff. This version restriction makes it so that tiff cannot be installed at the same time as byteorder 1.3. Until this issue is resolved, I will be preventing image (and thus tiff) from being available in the playground.

Thanks for listening!

birktj:
Fixes #21

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ description = "TIFF decoding and encoding library in pure Rust"
 repository = "https://github.com/PistonDevelopers/image-tiff"
 
 [dependencies]
-byteorder = "1.2.*"
-lzw = "0.10.*"
-num-derive = "0.2.*"
-num-traits = "0.2.*"
+byteorder = "1.2"
+lzw = "0.10"
+num-derive = "0.2"
+num-traits = "0.2"


bvssvni:
Merging.

bvssvni:
Thanks!

shepmaster:
Thank you!

PR #23, opened 2019-01-31
Tiff encoding

birktj:
This is a very much WIP implementation of tiff encoding. I believe that it is correct and can handle most cases, however the api is currently not very ergonomic or safe. It also lacks built-inn compression, however that kind of requires https://github.com/nwin/lzw/pull/1. 

Currently the api looks something like this, but I would very much like to improve it. 

```rust
let file = std::fs::File::create("test.tiff").unwrap();
let mut tiff = TiffEncoder::new_le(file);

tiff.write_header().unwrap();

tiff.new_ifd().unwrap();
tiff.new_ifd_entry(Tag::ImageWidth, &[100u32][..]);
tiff.new_ifd_entry(Tag::ImageLength, &[100u32][..]);
tiff.new_ifd_entry(Tag::BitsPerSample, &[8u16,8,8][..]);
tiff.new_ifd_entry(Tag::Compression, &[1u16][..]);
tiff.new_ifd_entry(Tag::PhotometricInterpretation, &[2u16][..]);
tiff.new_ifd_entry(Tag::SamplesPerPixel, &[3u16][..]);
tiff.new_ifd_entry(Tag::RowsPerStrip, &[100u32][..]);
tiff.new_ifd_entry(Tag::XResolution, &[Rational {n: 1, d: 1}][..]);
tiff.new_ifd_entry(Tag::YResolution, &[Rational {n: 1, d: 1}][..]);
tiff.new_ifd_entry(Tag::ResolutionUnit, &[1u16][..]);
// The two following tags are overwritten by `write_strip`
tiff.new_ifd_entry(Tag::StripOffsets, &[0u32][..]);
tiff.new_ifd_entry(Tag::StripByteCounts, &[0u32][..]);
tiff.finish_ifd().unwrap();

tiff.write_strip(&image_data).unwrap();
```

### DIFF
--- a/src/encoder/colortype.rs
+++ b/src/encoder/colortype.rs
@@ -1,9 +1,11 @@
+use crate::decoder::PhotometricInterpretation;
+
 /// Trait for different colortypes that can be encoded.
 pub trait ColorType {
-    /// The type of each sample of thes colortype
+    /// The type of each sample of this colortype
     type Inner: super::TiffValue;
     /// The value of the tiff tag `PhotometricInterpretation`
-    const TIFF_VALUE: u8;
+    const TIFF_VALUE: PhotometricInterpretation;
     /// The value of the tiff tag `BitsPerSample`
     fn bits_per_sample() -> Vec<u16>;
 }
@@ -11,7 +13,7 @@ pub trait ColorType {
 pub struct Gray8;
 impl ColorType for Gray8 {
     type Inner = u8;
-    const TIFF_VALUE: u8 = 1;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::BlackIsZero;
     fn bits_per_sample() -> Vec<u16> {
         vec![8]
     }
@@ -20,7 +22,7 @@ impl ColorType for Gray8 {
 pub struct Gray16;
 impl ColorType for Gray16 {
     type Inner = u16;
-    const TIFF_VALUE: u8 = 1;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::BlackIsZero;
     fn bits_per_sample() -> Vec<u16> {
         vec![16]
     }
@@ -29,7 +31,7 @@ impl ColorType for Gray16 {
 pub struct RGB8;
 impl ColorType for RGB8 {
     type Inner = u8;
-    const TIFF_VALUE: u8 = 2;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
     fn bits_per_sample() -> Vec<u16> {
         vec![8,8,8]
     }
@@ -38,7 +40,7 @@ impl ColorType for RGB8 {
 pub struct RGB16;
 impl ColorType for RGB16 {
     type Inner = u16;
-    const TIFF_VALUE: u8 = 2;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
     fn bits_per_sample() -> Vec<u16> {
         vec![16,16,16]
     }
@@ -47,7 +49,7 @@ impl ColorType for RGB16 {
 pub struct RGBA8;
 impl ColorType for RGBA8 {
     type Inner = u8;
-    const TIFF_VALUE: u8 = 2;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
     fn bits_per_sample() -> Vec<u16> {
         vec![8,8,8,8]
     }
@@ -56,7 +58,7 @@ impl ColorType for RGBA8 {
 pub struct RGBA16;
 impl ColorType for RGBA16 {
     type Inner = u16;
-    const TIFF_VALUE: u8 = 2;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
     fn bits_per_sample() -> Vec<u16> {
         vec![16,16,16,16]
     }
@@ -65,7 +67,7 @@ impl ColorType for RGBA16 {
 pub struct CMYK8;
 impl ColorType for CMYK8 {
     type Inner = u8;
-    const TIFF_VALUE: u8 = 5;
+    const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::CMYK;
     fn bits_per_sample() -> Vec<u16> {
         vec![8,8,8,8]
     }
--- a/src/encoder/colortype.rs
+++ b/src/encoder/colortype.rs
@@ -1,0 +1,68 @@
+pub trait ColorType {
+    type Inner: super::TiffValue;
+    const TIFF_VALUE: u8;
+    fn bits_per_sample() -> Vec<u16>;
+}
+
+pub struct Grey8;
+impl ColorType for Grey8 {
+    type Inner = u8;
+    const TIFF_VALUE: u8 = 1;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![8]
+    }
+}
+
+pub struct Grey16;
+impl ColorType for Grey16 {
+    type Inner = u16;
+    const TIFF_VALUE: u8 = 1;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![16]
+    }
+}
+
+pub struct RGB8;
+impl ColorType for RGB8 {
+    type Inner = u8;
+    const TIFF_VALUE: u8 = 2;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![8,8,8]
+    }
+}
+
+pub struct RGB16;
+impl ColorType for RGB16 {
+    type Inner = u16;
+    const TIFF_VALUE: u8 = 2;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![16,16,16]
+    }
+}
+
+pub struct RGBA8;
+impl ColorType for RGBA8 {
+    type Inner = u8;
+    const TIFF_VALUE: u8 = 2;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![8,8,8,8]
+    }
+}
+
+pub struct RGBA16;
+impl ColorType for RGBA16 {
+    type Inner = u16;
+    const TIFF_VALUE: u8 = 2;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![16,16,16,16]
+    }
+}
+
+pub struct CMYK8;
+impl ColorType for CMYK8 {
+    type Inner = u8;
+    const TIFF_VALUE: u8 = 5;
+    fn bits_per_sample() -> Vec<u16> {
+        vec![8,8,8,8]
+    }
+}
--- a/src/encoder/colortype.rs
+++ b/src/encoder/colortype.rs
@@ -7,68 +7,54 @@ pub trait ColorType {
     /// The value of the tiff tag `PhotometricInterpretation`
     const TIFF_VALUE: PhotometricInterpretation;
     /// The value of the tiff tag `BitsPerSample`
-    fn bits_per_sample() -> Vec<u16>;
+    const BITS_PER_SAMPLE: &'static [u16];
 }
 
 pub struct Gray8;
 impl ColorType for Gray8 {
     type Inner = u8;
     const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::BlackIsZero;
-    fn bits_per_sample() -> Vec<u16> {
-        vec![8]
-    }
+    const BITS_PER_SAMPLE: &'static [u16] = &[8];
 }
 
 pub struct Gray16;
 impl ColorType for Gray16 {
     type Inner = u16;
     const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::BlackIsZero;
-    fn bits_per_sample() -> Vec<u16> {
-        vec![16]
-    }
+    const BITS_PER_SAMPLE: &'static [u16] = &[16];
 }
 
 pub struct RGB8;
 impl ColorType for RGB8 {
     type Inner = u8;
     const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
-    fn bits_per_sample() -> Vec<u16> {
-        vec![8,8,8]
-    }
+    const BITS_PER_SAMPLE: &'static [u16] = &[8, 8, 8];
 }
 
 pub struct RGB16;
 impl ColorType for RGB16 {
     type Inner = u16;
     const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
-    fn bits_per_sample() -> Vec<u16> {
-        vec![16,16,16]
-    }
+    const BITS_PER_SAMPLE: &'static [u16] = &[16, 16, 16];
 }
 
 pub struct RGBA8;
 impl ColorType for RGBA8 {
     type Inner = u8;
     const TIFF_VALUE: PhotometricInterpretation = PhotometricInterpretation::RGB;
-    fn bits_per_sample() -> Vec<u16> {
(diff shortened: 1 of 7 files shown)


birktj:
I have now made some improvements to the encoder api and the above could now be written as:
```rust
let file = std::fs::File::create("test.tiff").unwrap();
let mut tiff = TiffEncoder::new(file).unwrap();

let mut image = tiff.new_image::<RGB8>(100, 100).unwrap();
    
let mut idx = 0;
while image.next_strip_sample_count() > 0 {
        let sample_count = image.next_strip_sample_count() as usize;
        image.write_strip(&image_data[idx..idx+sample_count]).unwrap();
        idx += sample_count;
}
image.finish().unwrap();
```

Currently I have created two apis, one low level and one semi-high level.

I think that after a simple clean-up and some documentation this patch should be ready.

Robzz:
Thank you for your work on this! I'll look a quick look at it now and one more when you think it's finished.

src/encoder/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    const BYTE_LEN: u32 = 1;
+    const FIELD_TYPE: u16 = 2;
+
+    fn count(&self) -> u32 {
+        self.len() as u32 + 1
+    }
+
+    fn write<W: Write>(&self, writer: &mut TiffWriter<W>) -> TiffResult<()> {
+        for x in self.chars() {
+            if x.is_ascii() {
```
Robzz: Would it make more sense to return an error if the string is not ASCII instead of silently skipping characters ? Doing so and checking the whole string at once, the function could be rewritten as something like:

```
if self.is_ascii() {
    writer.write_bytes(self.as_bytes())?;
    Ok(writer.write_u8(0)?)
}
else {
    Err(...)
}
```

birktj:
@Robzz I think I have fixed most of your concerns, however I have two concerns myself that I would like you to comment on before I am happy with the result.

The first is that I have used quite a bit of associated constants and the image crate has rust 1.24.1 as minimum supported version. I am not entirely sure if rust 1.24 supports them, should I simply use methods instead? 

On a related note I have used a lot of traits to control decoding. I think the `TiffValue` trait is a good idea, but I am not entirely sure if a `ColorType` trait is such a good idea?

If you do not think that these are big issues that I would say that this PR is finished.

Robzz:
Regarding associated constants, they were stabilized in rust 1.20 so compatibility with the image crate is ok. It might be a good idea to specify the version requirement in the README though.

I'll take a deeper look in the next few days with the TIFF spec my by side and play with it a bit.
I see you also added more tests, great!

birktj:
A few things i have noted when using this:
- Directly using a writer such as `File` is really slow, using 'BufWriter' helps a lot, but it would probably be better to optimize the writing. Optimizing writing strips of `u8` should be simple: remove `impl TiffValue for u8` and add a specialized `impl TiffValue for &[u8]`. However this won't help with strips of `u16` and other bitdepths. I was thinking that since the writer is writing using `NativeEdian` it should be safe to cast slices of different number types to slices of `u8` and write those using `write_all`. @Robzz what do you think of this?
- Currently the decoder doesn't support compression. I don't think this is a big issue and it should be simple to add a patch for this later, however maybe we should modify `new_image` to take another argument with an enum for compressions so this could be added in a non-breaking way later.
- Another reason that the lack of support for compression is interesting is that without compression it is possible to write images in a streaming manner without the `Seek` bound by simply precalculating where the ifd will end up. For some usecases this may be useful and we should maybe think of some sort of api for writing non-compressed images without the 'Seek' bound.

Robzz:
Hey, I still haven't looked at it in depth but I haven't forgotten, it's planned for this week end ;)

Adressing your points in order:

* Yeah, if writing values one by one then I'd expect writing without a BufWriter to be slow. Regarding encoding in `NativeEndian`, I don't see any reason why casting wouldn't work. Also the tiff spec says it's fine for an encoder to use whatever endianness it wants, so we're okay on this front.
* Sounds good, then we just need to return `TiffUnsupportedError::UnsupportedCompressionMethod` when a compression method is specified.
* I'll actually need to sit down and look at the code for that one.

src/encoder/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+/// With this type you can get a `DirectoryEncoder` or a `ImageEncoder`
+/// to encode tiff ifd directories with images.
+///
+/// See `DirectoryEncoder` and `ImageEncoder`.
+///
+/// # Examples
+/// ```
+/// # extern crate tempfile;
+/// # let mut file = tempfile::tempfile().unwrap();
+/// # let mut image_data = Vec::new();
```
Robzz: Since we don't really care about the values here, you can remove the loop below by just creating a Vec full of 0's like this: `vec![0; 100*100*3]`

src/encoder/colortype.rs
```diff
@@ -0,0 +1,72 @@
+/// Trait for different colortypes that can be encoded.
+pub trait ColorType {
```
Robzz: So, regarding how you weren't sure that this was a good idea. On one hand here's a bit of overlap between this and the `ColorType` enum at the crate root, but on the other hand having a trait with this `Inner` associated type is convenient because it allows functions like `write_image<C: ColorType>(&mut self, width: u32, height: u32, data: &[C::Inner]) -> TiffResult<()>` instead of having functions for every possible subpixel type. I would probably have ended up with something similar. In any case, I don't have a better idea right now, I'd say it's ok.

src/encoder/colortype.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// The type of each sample of thes colortype
```
Robzz: Small typo here

src/encoder/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+/// Type to represent tiff values of type `RATIONAL`
+pub struct Rational {
+    pub n: u32,
+    pub d: u32,
+}
+
+/// Trait for types that can be encoded in a tiff file
+pub trait TiffValue {
+    const BYTE_LEN: u32;
+    const FIELD_TYPE: u16;
```
Robzz: The decoder already has a type for this:  `decoder::ifd::Type`. It might be clearer to use that than hardcode `FIELD_TYPE` values in, it only lacks a way to retrieve the variant values but you can easily add a `ToPrimitive` derive since we already depend on `num-derive`.

src/encoder/mod.rs
```diff
@@ -0,0 +1,425 @@
...
+    rows_per_strip: u64,
+    strip_offsets: Vec<u32>,
+    strip_byte_count: Vec<u32>,
+    dropped: bool,
+    _phantom: std::marker::PhantomData<C>,
+}
+
+impl<'a, W: Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
+    fn new(mut encoder: DirectoryEncoder<'a, W>, width: u32, height: u32) -> TiffResult<ImageEncoder<'a, W, T>> {
+        let row_samples = width as u64 * <T>::bits_per_sample().len() as u64;
```
Robzz: There are a few remaining clippy warning regarding casts in this method and the next one.

Robzz:
I'm mostly done reading the code, looks good to me. I'll still have to play with it a bit, I'll let you know if I run into any issues.

src/encoder/mod.rs
birktj: You mean like this right?
```rust
const FIELD_TYPE: ifd::Type;
```
or do you mean removing `FIELD_TYPE` and add a `ifd::Type` parameters to the other functions?

src/encoder/colortype.rs
birktj: The only other thing that I could think of, would be to make `write_image` like this:
```rust
fn write_image<T: TiffValue>(&mut self, width: u32, height: u32, colortype: ColorType, data: T) -> TiffResult<()>
```
However you then loose some type safety and the user becomes sort of locked to what's supported by `ColorType`, so I prefer this solution.

src/encoder/mod.rs
birktj: Yeah, I guess that looks kind of strange. I think I was just copying from the test without thinking too much about it.

birktj:
On more thing that I have noticed is that there maybe should be some more convenience methods on `ImageEncoder` like `set_resolution`. And we could probably use those for compression with a `set_compression` as well, in which case we don't need to add an extra compression argument to any of the functions in this PR.

src/encoder/mod.rs
birktj: Will fix

src/encoder/mod.rs
Robzz: Yep i mean keeping `FIELD_TYPE` but with type `ifd::Type`. I think it makes the code slightly clearer to read (e.g. `Type::BYTE` instead of `1`) and it also makes sure that `ifd::Type` is the only place where we can possibly get the magic numbers wrong.

birktj:
@Robzz I have done some small fixes and I think all of your points are covered. Please let me know if there is anything more you would like me to do.

Robzz:
I reviewed your latest commit, everything looks good to me. I'm for merging this, problems we encounter we can fix later, an we'll certainly get more feedback from people trying to use this than you'll get from me in this PR. However, I don't have merge powers 'round here. @bvssvni (or anyone else with magic merge powers) could you please give it a look and merge if OK ?

src/encoder/colortype.rs
```diff
@@ -0,0 +1,74 @@
+use crate::decoder::PhotometricInterpretation;
+
+/// Trait for different colortypes that can be encoded.
+pub trait ColorType {
+    /// The type of each sample of this colortype
+    type Inner: super::TiffValue;
+    /// The value of the tiff tag `PhotometricInterpretation`
+    const TIFF_VALUE: PhotometricInterpretation;
+    /// The value of the tiff tag `BitsPerSample`
+    fn bits_per_sample() -> Vec<u16>;
```
HeroicKatora: I'm not sure why this is a function but the other is associated value is a constant. Could `const bits_per_sample: &'static [u16]` work as well?

src/encoder/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+    const BYTE_LEN: u32;
+    const FIELD_TYPE: ifd::Type;
+    fn count(&self) -> u32;
+    fn bytes(&self) -> u32 {
+        self.count() * Self::BYTE_LEN
+    }
+    fn write<W: Write>(&self, writer: &mut TiffWriter<W>) -> TiffResult<()>;
+}
+
+impl TiffValue for &[u8] {
```
HeroicKatora: Is there a reason why this is implemented for `&'_ [u8]` instead of `[u8]`? All required methods take `&self` and thus `Sized` is not required. The `impl` for `str` is as expected.

src/encoder/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+        // We write using nativeedian so this sould be safe
+        let slice = unsafe {
+            std::slice::from_raw_parts(self.as_ptr() as *const u8, self.len()*4)
+        };
+        writer.write_bytes(slice)?;
+        Ok(())
+    }
+}
+
+impl TiffValue for &[Rational] {
```
HeroicKatora: Same here, but it would also simplify the `for` loop to the intuitive `for x in self`.

src/encoder/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+impl TiffValue for str {
+    const BYTE_LEN: u32 = 1;
+    const FIELD_TYPE: ifd::Type = ifd::Type::ASCII;
+
+    fn count(&self) -> u32 {
+        self.len() as u32 + 1
+    }
+
+    fn write<W: Write>(&self, writer: &mut TiffWriter<W>) -> TiffResult<()> {
+        if self.is_ascii() {
```
HeroicKatora: The string `"\0"` fulfills `is_ascii()` but contains a null character. This is probably not intended?
https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=c4b602fed6fed80ce7346cc0b8fe3713

src/encoder/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+        Ok(DirectoryEncoder {
+            writer,
+            dropped: false,
+            ifd_pointer_pos,
+            ifd: BTreeMap::new(),
+        })
+    }
+
+    /// Write a single ifd tag.
+    pub fn write_tag<T: TiffValue>(&mut self, tag: Tag, value: T) {
```
HeroicKatora: Values could be passed by reference instead of value here to accomodate `T: ?Sized` and thus `impl TiffValue for [u8]`. Nothing in this method strictly requires the value property.

src/encoder/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+        // As per tiff spec each strip should be about 8k long
+        let rows_per_strip = (8000 + row_bytes - 1) / row_bytes;
+
+        let strip_count = (u64::from(height) + rows_per_strip - 1) / rows_per_strip;
+
+        encoder.write_tag(Tag::ImageWidth, width);
+        encoder.write_tag(Tag::ImageLength, height);
+        encoder.write_tag(Tag::Compression, 1u16);
+
+        encoder.write_tag(Tag::BitsPerSample, &<T>::bits_per_sample() as &[u16]);
```
HeroicKatora: Here is the superfluous allocation, `bits_per_sample() -> Vec<_>`. No cast, and passing the value by reference would shorten this to `write_tag(Tag::BitsPerSample, <T>::bits_per_sample);`

src/encoder/mod.rs
HeroicKatora: It would be possible to get the best of both worlds by adding an additional, generic impl:

```
impl<'a, T: TiffValue + ?Sized> TiffValue for &'a TiffValue {
    ...
}
```

Then neither this methods nor its callsites need changes but can benefit from the impl on `[u8]` directly.

src/encoder/mod.rs
```diff
@@ -0,0 +1,494 @@
...
+        }
+
+        let start_row = std::cmp::min(u64::from(self.height), self.strip_idx * self.rows_per_strip);
+        let end_row = std::cmp::min(u64::from(self.height), (self.strip_idx+1)*self.rows_per_strip);
+
+        (end_row - start_row) * self.row_samples
+    }
+
+    /// Write a single strip.
+    pub fn write_strip<'b>(&mut self, value: &'b [T::Inner]) -> TiffResult<()> where &'b [T::Inner]: TiffValue {
```
HeroicKatora: Implementing for `[u8]` etc also removes this explicit lifetime annotation.

src/encoder/writer.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    pub fn write_u64(&mut self, n: u64) -> Result<(), io::Error> {
+        self.writer.write_u64::<NativeEndian>(n)?;
+        self.offset += 8;
+
+        Ok(())
+    }
+
+    pub fn pad_word_boundary(&mut self) -> Result<(), io::Error> {
+        while self.offset % 4 != 0 {
```
HeroicKatora: Not a large issue, potentially solved by the compiler, for writers that have sideeffects or benefit from fewer write calls:

```
if self.offset % 4 != 0 {
    let align_len = 4 - (self.offset % 4);
    let align = &[0, 0, 0][..align_len];
    self.writer.write_all(align)?;
}
```

HeroicKatora:
There's nothing major, but the part about encoding of strings is potentially be problematic. Would it be too much of a problem to cleanup some code style along the way?

HeroicKatora:
And I can merge this as well, feel free to ping me on the weekend about it.

birktj:
@HeroicKatora I have fixed the string issue, and changed the impls to be on `[T]` instead of `&[T]`. Anything more you would like me too do?

HeroicKatora:
Thank you for your contribution!

PR #25, opened 2019-02-01
Test decoder output

### LINKED ISSUE #24: Better tests
I am working on  https://github.com/PistonDevelopers/image/issues/858, but I am not entirely confident that my modified decoder is correct. Some more thorough test that would check if the decoded data is correct would be nice. There should also be tests for things like horizontal predictor.

birktj:
These tests are based on the assumption that the current decoder is
correct and are to make sure that future changes to the decoder
wont break it.

This will fix #24

### DIFF
--- a/tests/decode_images.rs
+++ b/tests/decode_images.rs
@@ -11,8 +11,18 @@ fn test_gray_u8()
     let img_file = File::open("./tests/images/minisblack-1c-8b.tiff").expect("Cannot find test image!");
     let mut decoder = Decoder::new(img_file).expect("Cannot create decoder");
     assert_eq!(decoder.colortype().unwrap(), ColorType::Gray(8));
-    let img_res = decoder.read_image();
-    assert!(img_res.is_ok());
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Since this has been open for a long time now I suppose it is a matter of confidence in the change? But there is already another accepted merge based on it. I'm not sure but reading this with some known correct software (or convert with e.g. `imagemagick`) would also give groundtruths.

Robzz:
I just verified the values with the following quick'n dirty python opencv script:

```python
#!/usr/bin/env python3
import cv2
import numpy as np
from sys import argv

img = cv2.imread(argv[1], cv2.IMREAD_ANYCOLOR | cv2.IMREAD_ANYDEPTH)
try:
    rows, cols = img.shape
    print("Sum: ", np.sum(img))
except ValueError as e:
    rows, cols, channels = img.shape
    print("Sum: ", np.sum(img))
```

All 4 are OK :+1:

PR #26, opened 2019-02-01
Incremental decoding implemented

birktj:
This is an implementation of incremental decoding (decoding one strip at a time). This PR is the first step towards https://github.com/PistonDevelopers/image/issues/858. I am reasonably confident that the implementation is correct in the same cases as the last implementation. I have tested it on the tests from #25 and they are correct. 

The public api has changed in a small way with two new public methods on `Decoder`, namely `Decoder::strip_count` and `Decoder::read_strip`. If there are any objections to the api changes or the implementation I would be happy to sort them out.


birktj:
@fintelia (or anyone else) I wonder if you could give me any feedback on this PR so that I can continue with https://github.com/PistonDevelopers/image/issues/858?

fintelia:
I'm not really that familiar with the TIFF decoding logic, but assuming that tests pass and so forth the code looks reasonable. My one concern is that having `read_strip` allocate and return a Vector might cause you to have to make extra copies when implementing `Read`.

If you think this is ready, I can go ahead and merge this PR and we can address any rough edges in a future patch.

birktj:
@fintelia I tried to keep it compatible with the current `DecodingResult`. I believe this could be done better, but that would require some larger api changes which is something I am looking into for the longer run.

I guess if I marked `DecodingBuffer` and `read_strip_to_buffer` as public I could fix the extra copy issue and simply copy some of the `read_strip` logic into the eventual `Reader`. I will fix that now and then we should then have what we need to continue with an `TiffReader` in the image crate for now.

birktj:
@fintelia friendly ping. Would it be possible to merge this or do you have any other concerns.

Issue #30, opened 2019-02-27
Tests fail with "index out of bounds"


birktj:
Currently the tests fail with an index out of bounds error, this is probably a bug introduced in #26 that could only be discovered with the tests from #23. 

```
running 1 test
test decoder::stream::test::test_packbits ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/decode_images-564f01fb7ded3313

running 6 tests
test test_gray_u8 ... ok
test test_rgb_u8 ... ok
test test_string_tags ... ok
test test_gray_u16 ... ok
test test_decode_data ... ok
test test_rgb_u16 ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/encode_images-b9c0207f836596cd

running 5 tests
test test_gray_u8_roundtrip ... ok
test test_rgb_u8_roundtrip ... ok
test test_gray_u16_roundtrip ... ok
test encode_decode ... ok
test test_rgb_u16_roundtrip ... FAILED

failures:

---- test_rgb_u16_roundtrip stdout ----
thread 'test_rgb_u16_roundtrip' panicked at 'index out of bounds: the len is 19 but the index is 19', /rustc/9fda7c2237db910e41d6a712e9a2139b352e558b/src/libcore/slice/mod.rs:2463:10
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    test_rgb_u16_roundtrip

test result: FAILED. 4 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test encode_images'
```

birktj:
Looking a bit more into this, I found that the bug seems to come from two places, the first place is that in the test I try to read the image twice and that the decoder dosen't stop it from trying to read more strips than the image has.

I will try to come with a patch that fixes this and a few other bugs as soon as possible.

PR #36, opened 2019-03-01
Remove several panics in write_image

### LINKED ISSUE #35: Encoding panics if the input is too short
The following code causes a panic in the encoder:

```rust
let input_data = vec![1,2,3].as_slice();
let mut output = Vec::new();
let mut output_stream = Cursor::new(output);
if let Ok(mut tiff) = TiffEncoder::new(&mut output_stream) {
    tiff.write_image::<colortype::RGB8>(50, 50, &input_data)
}
```
Error message:
`thread 'main' panicked at 'index 7500 out of range for slice of length 173', src/libcore/slice/mod.rs:2349:5`

aschampion:
Closes #35. It's not decided this is the preferred behavior, but my opinion is that since `write_image` returns a result rather than `()`, the least surprising behavior is to panic in as few cases as possible. It's ambiguous whether `InvalidData` or `UnexpectedEof` would be more appropriate here.

The buffer length is compared vs `width * height` to avoid a redundant comparison for every strip (because Rust still doesn't have non-panicking range indexing 🙁), but this limits the `width * height` to `u32`, so would have to change if BigTIFF support is planned.

Two other unwraps are converted to trys. These should be uncontroversial.

### DIFF
--- a/src/encoder/mod.rs
+++ b/src/encoder/mod.rs
@@ -474,7 +474,11 @@ impl<'a, W: Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
     {
         // TODO: Compression
         let samples = self.next_strip_sample_count();
-        assert_eq!(value.len() as u64, samples);
+        if value.len() as u64 != samples {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "Slice is wrong size for strip").into());
+        }
 
         let offset = self.encoder.write_data(value)?;
         self.strip_offsets.push(offset as u32);
--- a/src/encoder/mod.rs
+++ b/src/encoder/mod.rs
@@ -247,13 +247,23 @@ impl<W: Write + Seek> TiffEncoder<W> {
     where
         [C::Inner]: TiffValue,
     {
+        let num_pix = (width as usize).checked_mul(height as usize)
+            .ok_or_else(|| std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "Image width * height exceeds usize"))?;
+        if data.len() < num_pix {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "Input data slice is undersized for provided dimensions").into());
+        }
+
         let encoder = DirectoryEncoder::new(&mut self.writer)?;
-        let mut image: ImageEncoder<W, C> = ImageEncoder::new(encoder, width, height).unwrap();
+        let mut image: ImageEncoder<W, C> = ImageEncoder::new(encoder, width, height)?;
 
         let mut idx = 0;
         while image.next_strip_sample_count() > 0 {
             let sample_count = image.next_strip_sample_count() as usize;
-            image.write_strip(&data[idx..idx + sample_count]).unwrap();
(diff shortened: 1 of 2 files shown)


src/encoder/mod.rs
```diff
@@ -238,13 +238,18 @@ impl<W: Write + Seek> TiffEncoder<W> {
 
     /// Convenience function to write an entire image from memory.
     pub fn write_image<C: ColorType>(&mut self, width: u32, height: u32, data: &[C::Inner]) -> TiffResult<()> where [C::Inner]: TiffValue {
+        if data.len() < (width * height) as usize {
```
Shnatsel: `width * height` could overflow `usize` on 32-bit systems. Please use [`checked_mul()`](https://doc.rust-lang.org/std/primitive.u32.html#method.checked_mul) instead.

HeroicKatora: Should be `(width as usize * height as usize)` or a checked mul.

HeroicKatora: See @Shnatsel

HeroicKatora: And convert `width` and `height` to `usize` before the multiplication.

aschampion: If the multiplication overflows (*without* first casting to `usize`), presumably an error should be returned because the image is too big for TIFF anyway (32 bit offsets)?

HeroicKatora: Not necessarily. Compression, less then 8 bits per color, ... Just keep it separate, the check is cheap either way.

aschampion: Fair enough, I pushed an amend so the check is done now after a `usize` cast.

Shnatsel:
CI test is failing due to #30

HeroicKatora:
> (because Rust still doesn't have non-panicking range indexing slightly_frowning_face),

Yes it does [slice::get](https://doc.rust-lang.org/std/primitive.slice.html#method.get) supports a generic `S: SliceIndex<[T]>` argument, implemented e.g. by `Range<usize>`.

aschampion:
> > (because Rust still doesn't have non-panicking range indexing slightly_frowning_face),
> 
> Yes it does [slice::get](https://doc.rust-lang.org/std/primitive.slice.html#method.get) supports a generic `S: SliceIndex<[T]>` argument, implemented e.g. by `Range<usize>`.

Awesome! I'd been following the RFC but somehow missed it getting stabilized so long ago.

HeroicKatora:
The `SliceIndex` trait itself is not stable to implement but its `impls` can be used in stable. Maybe that was the source of confusion.

birktj:
If this is the behavior we want, we should also fix `ImageEncoder::write_strip` that currently contains an `assert_eq`

aschampion:
Done.

aschampion:
Rebased on master to fix an unrelated failing test. All tests are green now.

Shnatsel:
This has once again diverged with master. Can we get a clear yay or nay on these changes from the maintainers so that this can be merged without any more divergences, or dropped if this direction is not desirable?

HeroicKatora:
Did the rebase right now, and a clear yes. Merging once travis says this is fine as well.

PR #32, opened 2019-02-28
Fix bugs, compiler warnings and add tests

birktj:
This PR adds a `Limits` struct to fix the memory exhaustion bug in #29, this changes the public API slightly, because `ifd::Entry::val` now also needs a `Limits` parameter. It adds some array bounds checking for indices that directly come from the tiff file. While I was at it I also fixed #13 by changing `Decoder::read_string` to return a `TiffResult` instead of a `FromUtf8Error` result. I also fixed some smaller compiler and clippy warnings. 

I added the samples from #28, #29 and #31 as new tests, thank you @Shnatsel!

This will fix #13, fix #28, fix #29, fix #30 and fix #31.

@Robzz and @HeroicKatora what do you think of this?

PR #37, opened 2019-03-28
Allow Decoder::next_image() to be called multiple times to allow paging through a multi-page TIFF

### LINKED ISSUE #34: Missing docs/examples for decoder.next_image()
The function `decoder.next_image()` is under-documented.

First off, it is not clear how this function relates to `read_image()` and what the difference between them is.

Second, the naive approach to using it does not compile:

```rust
fn decode(data: &[u8]) {
    let cursor = Cursor::new(data);
    if let Ok(mut decoder) = Decoder::new(cursor) {
        while decoder.more_images() {
            decoder.next_image();
        }
    }
}
```
This fails with the following error:
```
error[E0382]: borrow of moved value: `decoder`
  --> src/bin/tiff-afl-multi.rs:11:19
   |
11 |             while decoder.more_images() {
   |                   ^^^^^^^ value borrowed here after move
12 |                 decoder.next_image();
   |                 ------- value moved here, in previous iteration of loop
   |
   = note: move occurs because `decoder` has type `tiff::decoder::Decoder<std::io::Cursor<&[u8]>>`, which does not implement the `Copy` trait
```
A code example that uses it correctly would be appreciated.

jimrybarski:
Due to some ownership issues, it does not seem possible to iterate through a multi-page TIFF, since `next_image` can only be called once. I just fiddled with the borrows a bit. This should fix #34.


jimrybarski:
There is one failing test but it should be fixed by #32

birktj:
I want to merge this as it should make the library more usable. @HeroicKatora do you have any objections?

After this and #41 is merged I think we should release a new version of the library,

HeroicKatora:
Seems fine. If you're ready for releasing `0.3` feel free to ping me.

PR #39, opened 2019-04-17
Format code using 'cargo fmt'

Atul9:


birktj:
This is probably long overdue, thanks! I think we also should do a more thorough manually cleanup at some time. If there are no objections then I will merge this.

PR #42, opened 2019-04-25
Update readme with organization data

HeroicKatora:
Inserts some other badges that are typically present in the other
repositories.


HeroicKatora:
Ok, I think I double checked everything now.

PR #43, opened 2019-04-25
Release version 0.3

birktj:
With a lot of bug fixes and both encoding and incremental decoding implemented I think that it is time for a new version of this library. This will also allow work in the `image` library to implement these new features there. 

Could somebody from @image-rs/publish acknowledge and proceed here?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ authors = [
 license = "MIT"
 description = "TIFF decoding and encoding library in pure Rust"
 repository = "https://github.com/image-rs/image-tiff"
+exclude = ["tests/images/*", "tests/fuzz_images/*"]
 
 [dependencies]
 byteorder = "1.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tiff"
-version = "0.2.2"
+version = "0.3.0"
 authors = [
     "ccgn",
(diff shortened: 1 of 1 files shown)


HeroicKatora:
@birktj The fuzz_images are seemingly a problem for `crates.io`.

```
$ cargo publish --dry-run
    Updating crates.io index
   Packaging tiff v0.3.0 (/home/andreas/code/projects/image-tiff)
error: failed to prepare local package for uploading

Caused by:
  cannot package a filename with a special character `:`: tests/fuzz_images/divide_by_zero/id:000000,sig:06,src:000036,op:havoc,rep:4
```

HeroicKatora:
We should probably exclude that folder from the package content itself.

birktj:
Hmm, if you think that is a good solution then I support it. If not I could simply rename the images.

HeroicKatora:
Renaming the images will have to be done everytime new ones are generated as well, and is easy to forget. [We exclude the test and fuzz images in `image` as well](https://github.com/image-rs/image/blob/master/Cargo.toml#L21).

HeroicKatora:
Well, tbh, it isn't quite comparable since the `afl` fuzzer is its own crate so that the fuzzing corpus for `afl` is never part of the actual crate to begin with.

birktj:
@HeroicKatora I have excluded the images

HeroicKatora:
Published :rocket:

PR #44, opened 2019-05-03
Change minimal rustc version to 1.24.1

birktj:
The image crate has `rustc` version 1.24.1 as minimal supported version, however some of the newer PRs in this crates has used new rust features that is not supported by this verison. This PR adds `rustc` 1.24.1 as a travis target and fixes the parts which are not supported in rust 1.24.1

See also discussion in https://github.com/image-rs/image/pull/922

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: rust
 rust:
-  - stable
-  - beta
-  - nightly
+    - 1.24.1
+    - stable
+    - beta
+    - nightly
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -2,6 +2,7 @@
 
 use std::collections::HashMap;
 use std::io::{self, Read, Seek};
+use std::mem;
 
 use super::stream::{ByteOrder, EndianReader, SmartReader};
 use {TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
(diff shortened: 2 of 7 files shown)


HeroicKatora:
The reference pattern changes etc. are definitely going in the direction of getting rid of technical debt, imho. That tip my attitude in favour of updating the minimum `rustc` version in `image` rather than requiring it in all depending crates. This is now approaching ~100 reverted style improvements just for the sake of supporting an unsupport `rustc` version.

birktj:
Yeah, it was a bit more awkward than I expected. It turns out that quite a few new features has been added since 1.24.1

birktj:
Finally! That was a lot more work than expected. @HeroicKatora I think maybe we could release a version 0.3.1 of this library with this patch applied without merging it. Given that it is [5 months since the last major `image` version](https://github.com/image-rs/image/commit/7c6b7ed5d1ba5c2fd17883b49ed2eb0f8a34348e) I would propose to release a version with rust 1.24.1 supported this spring and then possibly a new version without support for rust 1.24.1 to the summer?

birktj:
Small overlook from my side there, sorry. Seems like the last version was in January, so it is not that bad.

HeroicKatora:
I wouldn't necessarily base a release of a *major* version on the time since the last *major* one, but releasing `0.3.1` with only this release and basing `image` on that *exact* version sounds possible. 

I'm only slightly confused if this requirement should be captured in SemVer, im gravitating towards yes but that would mean that we need at least *two* major versions of `image` until we have increased the compiler requirement: one for the release with `tiff = "=0.3.1"` and one for the compiler version bump. (For `tiff` we don't need another, I'm fine with declaring `0.3` as requiring a new compiler).

Let's give this a night of thought at least.

PR #46, opened 2019-07-10
CFA PhotometricInterpretation support

powturns:
Added support for the CFA (raw) PhotometricInterpretation type as per
the TIFF/EP spec

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -90,6 +90,7 @@ pub enum PhotometricInterpretation {
     CMYK = 5,
     YCbCr = 6,
     CIELab = 8,
+    CFA = 32803,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, FromPrimitive)]


HeroicKatora:
What's the impact of this PR other than the decoder failing with a slightly different `UnsupportedError` at a slightly different point in the code? If in fact the encoded value with which [this line](https://github.com/image-rs/image-tiff/blob/c9675360cffef5cb3b66e1163a1f8decd77874e8/src/decoder/mod.rs#L335) fails in the decoder is important, then I would rather the error type is expanded to permit inspecting the raw `u32` rather than special case single possible values that are not *actually* supported anywhere else in the library.

If I understand this correctly, then `PhotometricInterpretation` was specified to be extensible. As such, there may be other options that could be added. Such a choice comes up quite often in networking as well and I found `smoltcp` deals with this quite well using a custom wrapper macro to define enums to have an additional 'unknown' fallback option. [See here for definition](https://github.com/m-labs/smoltcp/blob/1387d4011c8c3dd7cac1f7100cc20c8f3d36bb59/src/macros.rs#L26-L77) and [here for usage](https://github.com/m-labs/smoltcp/blob/1387d4011c8c3dd7cac1f7100cc20c8f3d36bb59/src/wire/ethernet.rs#L6-L13). The only thing is that there should be an `__Nonexhaustive` variant to allow additions in a non-breaking change.

If the goal is purely to document some possible options then at least a publicly accessible point-of-reference would have been useful, so that actual interoperability can be added in the future. (I suppose #47 contains some background information).

51systems:
It seems I was a bit too hasty with my PR.
Since `PhotometricInterpretation` `CFA` is in the EP spec, I don't know if its necessary to go the 'unknown' route as opposed to including it in the enum; however, I can modify the [match](https://github.com/image-rs/image-tiff/blob/c9675360cffef5cb3b66e1163a1f8decd77874e8/src/decoder/mod.rs#L335) statement to fail on the `PhotometricInterpretation::CFA` type.

PR #49, opened 2019-08-12
Old-style DEFLATE compression support

Masterchef365:
Has not been tested on many images, but seems to work with ones I have. Only enabled for the 0x80B2 magic, but may also work for 0x0008.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ byteorder = "1.2"
 lzw = "0.10"
 num-derive = "0.3"
 num-traits = "0.2"
-libflate = "0.1" # TODO: Put behind a feature?
+miniz_oxide = "0.3" # TODO: Put behind a feature?
 
 [dev-dependencies]
 tempfile = "3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ byteorder = "1.2"
 lzw = "0.10"
 num-derive = "0.3"
 num-traits = "0.2"
+libflate = "0.1" # TODO: Put behind a feature?
 
 [dev-dependencies]
 tempfile = "3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ byteorder = "1.2"
 lzw = "0.10"
 num-derive = "0.3"
 num-traits = "0.2"
-miniz_oxide = "0.3" # TODO: Put behind a feature?
+miniz_oxide = "0.3"
 
 [dev-dependencies]
 tempfile = "3.0"
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -1,9 +1,10 @@
 //! All IO functionality needed for TIFF decoding
 
+use crate::error::TiffResult;
 use byteorder::{BigEndian, LittleEndian, ReadBytesExt};
 use lzw;
-use std::io::{self, Read, Seek};
 use miniz_oxide::inflate;
+use std::io::{self, Read, Seek};
 
 /// Byte order of the TIFF file.
 #[derive(Clone, Copy, Debug)]
@@ -75,15 +76,12 @@ pub struct DeflateReader {
 }
 
 impl DeflateReader {
-    pub fn new<R: Read + Seek>(
-        reader: &mut SmartReader<R>,
-    ) -> io::Result<(usize, Self)> {
+    pub fn new<R: Read + Seek>(reader: &mut SmartReader<R>) -> TiffResult<(usize, Self)> {
         let byte_order = reader.byte_order;
         let mut compressed = Vec::new();
         reader.read_to_end(&mut compressed)?;
-        let uncompressed = inflate::decompress_to_vec_zlib(&compressed).map_err(|error| {
-            std::io::Error::new(std::io::ErrorKind::InvalidData, format!("DEFLATE error: {:?}", error))
-        })?;
+        let uncompressed = inflate::decompress_to_vec_zlib(&compressed)?;
+
         Ok((
             uncompressed.len(),
             Self {
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -3,6 +3,7 @@
 use byteorder::{BigEndian, LittleEndian, ReadBytesExt};
 use lzw;
 use std::io::{self, Read, Seek};
+use libflate::zlib;
 
 /// Byte order of the TIFF file.
 #[derive(Clone, Copy, Debug)]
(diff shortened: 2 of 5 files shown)


HeroicKatora:
One thing to note here is the choice for the deflation library: `miniz_oxide` is the successor to `deflate`, and `libflate` only compares against the latter in their benchmarks. `miniz_oxide` might be faster?

Since it is an internal implementation detail and there are no publicly exposed interfaces that depend on the choice for the inflation library this is not critical but if you have the time I'd be interested in an evaluation nevertheless. Partially because this is also relevant to [a similar issue in `png`](https://github.com/image-rs/image-png/issues/151).

HeroicKatora:
Note: compile failure is due to a CI assertion for compatibility with Rust `1.24.1`. This may no longer be as relevant as it was in the past and we may bump it to `1.34`. `miniz_oxide` may require this while `libdeflate` seems more compatible with older versions.

Masterchef365:
Funny enough, I actually started looking at `miniz_oxide` for my own project after I wrote this PR. I just tested it on my own data and it was _noticeably_ faster, so I went ahead and merged my miniz branch into the deflate one. This could still use some more optimization (Maybe streaming decompression) but this at least functions for now.

HeroicKatora:
@Masterchef365  Ping, you should be able to use `miniz_oxide` now as `rustc` version has been bumped there is no longer a reason to feature gate this.

Masterchef365:
@HeroicKatora Rebased and pushed, looks like tests pass now.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ byteorder = "1.2"
 lzw = "0.10"
 num-derive = "0.3"
 num-traits = "0.2"
+miniz_oxide = "0.3" # TODO: Put behind a feature?
```
HeroicKatora: No longer `TODO`.

src/decoder/stream.rs
```diff
@@ -67,6 +68,44 @@ pub trait EndianReader: Read {
...
+/// Reader that decompresses DEFLATE streams
+pub struct DeflateReader {
+    buffer: io::Cursor<Vec<u8>>,
+    byte_order: ByteOrder,
+}
+
+impl DeflateReader {
+    pub fn new<R: Read + Seek>(
+        reader: &mut SmartReader<R>,
+    ) -> io::Result<(usize, Self)> {
```
HeroicKatora: `io::Error` is suboptimal here, prefer a `TiffError` here. Then you can differentiate between `reader` failing and and the `inflate` failing. The latter should have proper diagnosis with a new variant in `TiffFormatError` since I don't think any of the existing ones fit. The error variant should have an opaque type encapsulating the details from `miniz_oxide`.

src/decoder/stream.rs
```diff
@@ -67,6 +68,44 @@ pub trait EndianReader: Read {
...
+        let byte_order = reader.byte_order;
+        let mut compressed = Vec::new();
+        reader.read_to_end(&mut compressed)?;
+        let uncompressed = inflate::decompress_to_vec_zlib(&compressed).map_err(|error| {
```
HeroicKatora: Can this read or decompression subvert the byte limitation from `Limits`?

src/decoder/stream.rs
Masterchef365: Sounds good, would it be inappropriate to just add a variant named InflateError encapsulating the TINFLStatus miniz returns on error into TiffFormatError? Or are you asking for a more generalized custom enum that would hold constant even if we switched to a different DEFLATE implementation later?

HeroicKatora: Something where the interface doesn't change if `miniz_oxide` is replaced would be optimal. However, the internal representation need not be converted which would optimally make the `Display` and `Debug` implementation informative.

```
enum TiffFormatError {
    InflateError(InflateError),
    …
}

pub struct InflateError {
    status: TINFLStatus,
}
```

HeroicKatora: Actually, `TINFLStatus` doesn't add a `Display` implementation so it may be better to not print the details for display.

Masterchef365: Yeah I guess it could just have the display implementation be "Failed to decode INFLATE data" or something

HeroicKatora: No all-caps, best to think of display implementations as a standard sentence with Rust capitalization rules.

src/decoder/stream.rs
Masterchef365: It can, but the only way I can forsee fixing this (and increase performance) is by using streaming decompression. In the case of LZW this is supported naturally, but miniz_oxide's documentation currently states that "As of now this (streaming) is mainly inteded for use to build a higher-level wrapper" (referring to decompress_to_vec()). Ideally, streaming would receive first-class support upstream, but until that happens It could be implemented within image-tiff.

HeroicKatora: We could at least ensure that the compressed representation does not exceed it. [The implementation does not appear to be magic](https://docs.rs/miniz_oxide/0.3.2/src/miniz_oxide/inflate/mod.rs.html#73) so maybe it's possible to upstream a wrapper or have our own temporarily.

src/error.rs
```diff
@@ -55,10 +57,28 @@ impl fmt::Display for TiffFormatError {
...
         }
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct InflateError {
+    status: TINFLStatus,
+}
+
+impl From<TINFLStatus> for TiffFormatError {
```
HeroicKatora: The implementations of `From<_>` are part of the public interface. These two should be private inherent methods instead.

Masterchef365: Hmm, maybe it's not worth adding methods for then. While this is a bit wordy, it only happens in one spot:
```rust
let uncompressed = inflate::decompress_to_vec_zlib(&compressed).map_err(|status| {
    TiffError::FormatError(TiffFormatError::InflateError(InflateError::new(status)))
})?;
```
Thoughts?

src/decoder/stream.rs
Masterchef365: Agreed, I suppose it would be prompt to open a separate issue until the temporary custom solution is written or changes are made upstream to miniz_oxide? For now, ensuring the compressed representation is under the size restriction is a good idea and I'll go ahead and implement it.

src/error.rs
HeroicKatora: I'd still prefer a method since it doesn't capture any locals. And it takes as much visual space as the current `From<_>` impl so it shouldn't be more confusing.

Shnatsel:
Looks like the comments were resolved. Is there anything else that needs to be done before merging this?

HeroicKatora:
I haven't had time to review the last commit which addresses [this](https://github.com/image-rs/image-tiff/pull/49#discussion_r330857518).

PR #51, opened 2019-08-15
Signed value support

51systems:
Add support for SBYTE, SSHORT, SLONG, SRATIONAL as per the Tiff 6.0 spec
Add support for multiple BYTE

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -7,7 +7,7 @@ use std::mem;
 use super::stream::{ByteOrder, EndianReader, SmartReader};
 use {TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
 
-use self::Value::{Ascii, List, Rational, Unsigned};
+use self::Value::{Ascii, List, Rational, Unsigned, Signed, SRational};
 
 macro_rules! tags {
     {$(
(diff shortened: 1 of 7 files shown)


src/decoder/ifd.rs
```diff
@@ -166,14 +201,51 @@ impl Entry {
...
             // at a different endianess of file/computer.
             (Type::BYTE, 1) => Ok(Unsigned(u32::from(self.offset[0]))),
+            (Type::BYTE, 2) => offset_to_bytes(2, self),
+            (Type::BYTE, 3) => offset_to_bytes(3, self),
+            (Type::BYTE, 4) => offset_to_bytes(4, self),
+            (Type::BYTE, n) => {
+                if n as usize > limits.decoding_buffer_size / mem::size_of::<Value>() {
+                    return Err(TiffError::LimitsExceeded);
+                }
+                let mut v = Vec::with_capacity(n as usize);
```
HeroicKatora: Now would be a good opportunity to factor out these 4 lines, reused across many entry types, into a dedicated method.

51systems: Done via closure

birktj:
Looks good to me

HeroicKatora:
Great work, thank you. You seem to be invested in `tiff`. Want to join `image-rs`? You can be invited to the `image-rs` collaborators team, which allows you access to this and the other repos if you want. See [the contribution policy](https://github.com/image-rs/organization/blob/master/CONTRIBUTING.md) and [code of conduct](https://github.com/image-rs/organization/blob/master/CODE_OF_CONDUCT.md).

51systems:
That would be great! I'm still pretty new to both rust and image processing, but I'd like to contribute the dng encoder if it ever gets to an acceptable state.

PR #52, opened 2019-08-16
Make fields on Limits public

Masterchef365:
Decoder has a public constructor that takes an argument of Limits, but there is no constructor for the struct. This PR makes the fields on Limits publicly assignable, so that you may increase the limits if you are reading gigantic TIFF files.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -129,10 +129,10 @@ pub struct Limits {
     /// 256MiB. If the entire image is decoded at once, then this will
     /// be the maximum size of the image. If it is decoded one strip at a
     /// time, this will be the maximum size of a strip.
-    decoding_buffer_size: usize,
+    pub decoding_buffer_size: usize,
     /// The maximum size of any ifd value in bytes, the default is
     /// 1MiB.
-    ifd_value_size: usize,
+    pub ifd_value_size: usize,
 }
 
 impl Default for Limits {


HeroicKatora:
This makes all fields of the structure public, which means that adding more more fields would be a breaking change. Not sure we want to commit to this but it would be very useful to be able to change the limits.

Masterchef365:
Thought about that, but considered that maybe the Limits struct wouldn't change - in which case just assigning fields directly would simplify code a little. How about postfix constructors instead? For example you could do:
```
let limits = Limits::default()
    .with_max_buffer_size(256 * 1024 * 1024)
    .with_max_ifd_value_size(1024 * 1024 * 1024);
let decoder = decoder.with_limits(limits);
```
Or, if we were okay with a change that doesn't break anyone's code right now (Since noone has been able to use use the Limits struct so far)
```
let decoder = Decoder::new(reader)
    .with_max_buffer_size(256 * 1024 * 1024)
    .with_max_ifd_value_size(1024 * 1024 * 1024);
```
That way if more limits were added in the future users wouldn't have to change code at all.

HeroicKatora:
It would also be possible to use a trick of introducing a private fields. This form is noted as a current, hacky way and motiviation for [the `non_exhaustive` nightly feature and RFC](https://github.com/rust-lang/rfcs/blob/master/text/2008-non-exhaustive.md#structs).

```

/// Decoding limits
#[derive(Clone, Debug)]
pub struct Limits {
    /// The maximum size of any `DecodingResult` in bytes, the default is
    /// 256MiB. If the entire image is decoded at once, then this will
    /// be the maximum size of the image. If it is decoded one strip at a
    /// time, this will be the maximum size of a strip.
    decoding_buffer_size: usize,
    pub decoding_buffer_size: usize,
    /// The maximum size of any ifd value in bytes, the default is
    /// 1MiB.
    ifd_value_size: usize,
    pub ifd_value_size: usize,
    /// Private field to allow adding more options. This forces matching to have a trailing `..` 
    //// for the additional privates field that can not be named.
    non_exhaustive: (),
}
```

We don't have the constructor problem since `Default` is available in any case and should be used. You'd be expected to customize limits you know about but keep others unchanged.

Masterchef365:
I went ahead sketched up and tested an implementation of what I described in my repo [here](https://github.com/Masterchef365/image-tiff/tree/postfix_decoder_constructor_limits). If you decide the `non_exhaustive: ()` method is better, I'd be happy to implement and document it though.

HeroicKatora:
It would be less intrusive and I'm not sure what extra functionality you get from the other implementation that couldn't also be added without moving `Limits` fully into `Decoder`? You could still add the postfix builder methods `with_decoding_buffer_size` and `with_ifd_value_size` without that change. A getter for the current limits of a `decoder` would also make `with_limits` more helpful so that it can provide similar functionality.

Masterchef365:
I had removed Limits in that branch because it seemed somewhat unnecessary given that it was no longer passed in, but I do understand that it adds some segmentation to the code. So you're thinking keep the Limits struct, but also allow for postfix? So all of the following would be possible:
```rust
let limits = Limits {
    decoding_buffer_size: 256 * 1024 * 1024,
    ifd_value_size: 1024 * 1024 * 1024,
    ..Default::default()
};
let decoder = Decoder::new(reader).with_limits(limits);
let decoder = Decoder::new(reader)
    .with_max_buffer_size(256 * 1024 * 1024)
    .with_max_ifd_value_size(1024 * 1024 * 1024);
let decoding_buffer_size = decoder.limits().decoding_buffer_size;
```
I do think that having both the limits struct and the postfix would be a little bit redundant though.

Another option I can think of is this:
```rust
let limits = Limits::defaults()
    .with_decoding_buffer_size(256 * 1024 * 1024)
    .with_ifd_value_size(1024 * 1024 * 1024);
let decoder = Decoder::with_limits(limits);
let limits: &Limits = decoder.limits();
```
Which would avoid the `non_exhaustive: ()` field.

HeroicKatora:
Superseeded by #55

PR #55, opened 2019-09-22
Make Limits struct fields publicly accessible

trolleyman:
Following on from https://github.com/image-rs/image-tiff/pull/52, however this is a more conservative change that I hope will get accepted, as I need to be able to open large TIFF files.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -129,10 +129,14 @@ pub struct Limits {
     /// 256MiB. If the entire image is decoded at once, then this will
     /// be the maximum size of the image. If it is decoded one strip at a
     /// time, this will be the maximum size of a strip.
-    decoding_buffer_size: usize,
+    pub decoding_buffer_size: usize,
     /// The maximum size of any ifd value in bytes, the default is
     /// 1MiB.
-    ifd_value_size: usize,
+    pub ifd_value_size: usize,
+    /// The purpose of this is to prevent all the fields of the struct from
+    /// being public, as this would make adding new fields a major version
+    /// bump.
+    _non_exhaustive: (),
 }
 
 impl Default for Limits {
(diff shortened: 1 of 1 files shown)


HeroicKatora:
Had not followed up on this as I saw the broken build. I had not realized that this is some dev-dependency that no longer builds with `rustc = 1.24`? :disappointed: It's at least not the main build that is broken.

src/decoder/mod.rs
```diff
@@ -129,17 +129,22 @@ pub struct Limits {
...
-    decoding_buffer_size: usize,
+    pub decoding_buffer_size: usize,
     /// The maximum size of any ifd value in bytes, the default is
     /// 1MiB.
-    ifd_value_size: usize,
+    pub ifd_value_size: usize,
+    /// The purpose of this is to prevent all the fields of the struct from
+    /// being public, as this would make adding new fields a major version
+    /// bump.
+    _non_exhaustive: (),
```
birktj: It is probably better to just use `#[doc(hidden)]`.

birktj:
I think we should be able to merge this. The reason for the `Limits` struct ended up in this state was simply an oversight and I think this is an acceptable solution. The `rustc` version requirement was updated in https://github.com/image-rs/image-tiff/commit/ce1d4cb289873eb8e780105902ec20819749a692 so that shouldn't be a problem anymore.

src/decoder/mod.rs
HeroicKatora: Are there semantic difference to consider? The enum solutions use `doc(hidden)` since there is no such thing as a private variant so we can't really compare it directly. Is there prior art?

HeroicKatora: This is the solution suggested in the relating rfc: https://rust-lang.github.io/rfcs/2008-non-exhaustive.html

birktj: Of course, this is a `struct`, not an `enum`. :sweat_smile:

HeroicKatora:
Good, agree on the merge. Thank you, @trolleyman.

PR #56, opened 2019-09-22
Renamed test files

trolleyman:
The file names of some tests in the `tests/` folder were causing some failures on Windows, as they contain a `:` character. I replaced all that I could find to an equals symbol instead. Hopefully it's not a major issue. Also, I checked that the tests all passed correctly still.

The failure:
image

The tests passing:
image


trolleyman:
Cheers!

PR #58, opened 2019-10-01
update num-derive to 0.3

paolobarbolini:
In an effort to reduce duplicate dependency versions in projects, update `num-derive` to 0.3, which upgrades `proc-macro2`, `quote` and `syn` to version 1

For the updated num-derive to work the minimal rustc version becomes 1.32.0. We bump it to 1.34.2 for consistency and to accomodate other changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-    - 1.24.1
+    - 1.34.2
     - stable
     - beta
     - nightly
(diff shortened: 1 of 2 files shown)


JohnTitor:
I think it's reasonable to bump up the minimum supported version but I'd like to hear other's opinions. What about? @HeroicKatora

HeroicKatora:
Can you bump to `1.34.2` instead? The choice is not arbitrary and not descriptive, it ensures `image = 0.21` builds on older or slower platforms. If we are to publish a new major version then that is at least the correct compiler version.

HeroicKatora:
Fine. Since there is `deflate` also basically blocked on the `rustc` version I'm very much for bumping version right now.

PR #62, opened 2019-11-08
Multipage image patch

onnotechno:
While using image-tiff for multipage image I ran into 2 things:
1. the ifd offset is written twice for each image, which results into that the ifd offset of the second and later images was 0 and followed by the correct location.
2. the returned offset of the TiffWriter was 2 bytes off, starting at the second page. The easiest way to resolve this was by using the seek(SeekFrom::Current(0)) functionality of the writer object, instead of fixing the offset housekeeping.

### DIFF
--- a/src/encoder/writer.rs
+++ b/src/encoder/writer.rs
@@ -10,6 +10,7 @@ impl TiffByteOrder for LittleEndian {
     fn write_header<W: Write>(writer: &mut TiffWriter<W>) -> TiffResult<()> {
         writer.writer.write_u16::<LittleEndian>(0x4949)?;
         writer.writer.write_u16::<LittleEndian>(42)?;
+        writer.offset += 4;
 
         Ok(())
     }
@@ -19,6 +20,7 @@ impl TiffByteOrder for BigEndian {
     fn write_header<W: Write>(writer: &mut TiffWriter<W>) -> TiffResult<()> {
         writer.writer.write_u16::<BigEndian>(0x4d4d)?;
         writer.writer.write_u16::<BigEndian>(42)?;
+        writer.offset += 4;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


src/encoder/writer.rs
```diff
@@ -20,83 +19,73 @@ impl TiffByteOrder for BigEndian {
...
         Ok(())
     }
 
+}
+
+impl<W: Write + Seek> TiffWriter<W> {
     pub fn pad_word_boundary(&mut self) -> Result<(), io::Error> {
-        if self.offset % 4 != 0 {
+        let offset = self.offset();
+        if offset % 4 != 0 {
```
HeroicKatora: This is not equivalent to the previous code as it nows aligns to the offset of the underlying writer instead of the offset within the tiff. Maybe this is connected to the cause of the issue? I do not understand enough about this specific part of the Tiff specification to say but your PR comment does not reference this in as a root cause.

src/encoder/writer.rs
```diff
@@ -20,83 +19,73 @@ impl TiffByteOrder for BigEndian {
...
     pub fn write_u64(&mut self, n: u64) -> Result<(), io::Error> {
         self.writer.write_u64::<NativeEndian>(n)?;
-        self.offset += 8;
 
         Ok(())
     }
 
+}
+
+impl<W: Write + Seek> TiffWriter<W> {
```
HeroicKatora: Introducing the extra bound here is not a breaking change because `TiffWriter` is private. Right?

HeroicKatora:
Thanks for the PR, it seems the current test suite does not cover multi-page tiff encoding extensively enough. Would it be possible to add some test cases as well, maybe ones that demonstrate the current issue and its fix here?

src/encoder/writer.rs
birktj: That is correct

src/encoder/writer.rs
birktj: I think this is more correct. When I think about it, the previous code would have a problem when the writer does not start on offset 0. This is because the code sets the starting offset to 0 and then uses `seek(SeekFrom::Start())` which would then be incorrect. This should be fixable by changing the constructor to check the offset or use `SeekFrom::Current` instead. Not sure if it would fix this issue though.

src/encoder/writer.rs
```diff
@@ -106,8 +95,11 @@ impl<W: Write> TiffWriter<W> {
 
 impl<W: Seek> TiffWriter<W> {
     pub fn goto_offset(&mut self, offset: u64) -> Result<(), io::Error> {
-        self.offset = offset;
         self.writer.seek(SeekFrom::Start(offset as u64))?;
         Ok(())
     }
+
+    pub fn offset(&mut self) -> u64 {
+        self.writer.seek(SeekFrom::Current(0)).unwrap()
```
birktj: I think this method should return `Result<u64>` instead of using `unwrap`

src/encoder/writer.rs
HeroicKatora: The previous code was mistaken in its implementation of `goto_offset`. But the scope of the PR is way larger than that and does not only address that. And I fear that this also introduced *another* subtle bug here despite trying to correct seeking issues.

HeroicKatora: Specifically, the root cause seems to be located here https://github.com/image-rs/image-tiff/blob/6aea7b33c1d4c0b0c7384588513313ae9a52ae59/src/encoder/writer.rs#L108-L111

Note that `offset` should be relative to some stored starting offset. This PR abandons all of the offset tracking for fallible I/O operations (which are more expensive sys calls depending on the underlying writer) without sufficient motivation. And these changes in total affect more than the bug it lists. Is there some other bug I haven't seen? It does not explain.

birktj: I also agree that the manual offset tracking is better, which is the reason I wrote it that way to begin with. 

This PR also adds some changes to the IFD offset writing. I believe (though not entirely convinced) that these changes are correct. The problem that these changes address is what causes problems with multi-image tiff.

onnotechno:
I found the cause of the invalid offset while using multi page images. It was caused by not increasing the internal offset while adding padding.

src/encoder/writer.rs
```diff
@@ -98,6 +98,7 @@ impl<W: Write> TiffWriter<W> {
             let padding = [0, 0, 0];
             let padd_len = 4 - (self.offset % 4);
             self.writer.write_all(&padding[..padd_len as usize])?;
+            self.offset += padd_len;
```
birktj: That was a very nice catch!

birktj:
If you could fix the seeking like discussed as well, then I think this PR is ready to be merged.

onnotechno:
@birktj so adding the the start_offset to the SeekFrom::Current implementation? I will add a multipage test to the testcases

birktj:
Yes, that one should be good.

onnotechno:
A mulitpage image testcases has been added to the tests. This testcase fails with the master and succeeds with this patched version.

I have been looking into the start_offset modification, and that seems not to be possible. `TiffWriter::new` has to call `writer.seek(SeekFrom::Current(0)).unwrap()` to acquire the start_offset, so `writer` must implement the `Seek` trait when `TiffWriter::new` is being called.
In the functions `DirectoryEncoder::write_tag` and `DirectoryEncoder::write_directory` the `TiffWriter::new` is used with a Vector and a slice, and both of them do not implement the `Seek` trait.

It is probably best to only apply the `self.offset += padd_len;` patch to the function `TiffWriter::pad_word_boundary`. This fixes the invalid offset issue.

HeroicKatora:
> I have been looking into the start_offset modification, and that seems not to be possible. TiffWriter::new has to call writer.seek(SeekFrom::Current(0)).unwrap() to acquire the start_offset, so writer must implement the Seek trait when TiffWriter::new is being called.
In the functions DirectoryEncoder::write_tag and DirectoryEncoder::write_directory the TiffWriter::new is used with a Vector and a slice, and both of them do not implement the Seek trait.

The alternative `io::Cursor<Vec<u8>>` implements `Seek` and anyways this private part of the implementation could be changed to allow for this call, such as creating a new constructor `new_with_offset` and otherwise assuming offset of `0`. (Don't hastily declare it 'not possible' :smile:). However, that need not happen in this PR imo.

tests/encode_images.rs
```diff
@@ -260,4 +260,35 @@ fn test_signed() {
...
         assert_eq!(decoder.get_tag(Tag::Unknown(65031)).unwrap().into_i32_vec().unwrap(), [-1_i32, 100, 2, 100]);
     }
-}
\ No newline at end of file
+}
+
+#[test]
+/// check multipage image handling
+fn test_multipage_image() {
+    let mut img_file = tempfile::tempfile().unwrap();
```
HeroicKatora: This should not need file io.

```suggestion
    let mut img_file = tempfile::tempfile().unwrap();
```

HeroicKatora: But other examples already use `tempfile` so it doesn't matter greatly.

HeroicKatora:
Thanks for the contribution!

Issue #57, opened 2019-09-27
`num-derive` dependency worth it?


HeroicKatora:
The crate currently uses `num-derive` to add some `num-traits::FromPrimitive` implementations to a couple of enumerations. Since this is done via a procmacro it requires sequencing in the compilation pipeline, compiling `proc-macro2`/`syn` prior. All cases are simple enumerations such as :

```rust
pub enum Type {
    BYTE = 1,
    ASCII = 2,
    SHORT = 3,
    LONG = 4,
    RATIONAL = 5,
}
```

The functionality could also be added with a standard macro in such simple cases, see [this example in `smoltcp`](https://github.com/m-labs/smoltcp/blob/master/src/macros.rs#L26-L77) which adds very similar converters for `std::convert::{From, To}`.

birktj:
It is probably not necessary, for example we [already use a macro to create the `Tags` enum](https://github.com/image-rs/image-tiff/blob/master/src/decoder/ifd.rs#L12-L38). A new macro that adds `From` and `To` impls should be fine.

Robzz:
Yeah it's totally droppable. `image` used to depend on it, but only because of the tiff code, and I [removed it from image](https://github.com/image-rs/image/pull/783/files#diff-b4aea3e418ccdb71239b96952d9cddb6) when extracting the code into this crate. It's probably worth doing.

HeroicKatora:
One remaining question is one of style. Either go with the `Unknown(_)` variant from `Tag` or make the conversion return an `Option` like it does right now in the [`FromPrimitive`](https://docs.rs/num-traits/0.2.8/num_traits/cast/trait.FromPrimitive.html#tymethod.from_i64) trait.

Issue #59, opened 2019-10-02
`CompressionMethod` and `PhotometricInterpretation` should be NonExhaustive


HeroicKatora:
This is a small compatibility hazard, the only enum variants are the currently supported or introduced variants for these two (and potentially other) fields in `tiff`. Adding new variants breaks user code that matches on them, it would be customary to use instead an explicit `__NonExhaustive` variant that is not publicly documented and explicitely should not be matched against.

HeroicKatora:
`Tag` could be non exhaustive as well. Adding detection for now tags, effectively moving from `Unknown` to a new known variant such as potentially suggested in #47, does not require a breaking change.

PR #63, opened 2019-11-10
Replace FromPrimitive with our own tag macro

HeroicKatora:
This is more flexible and removes the dependency.

Note that `FromPrimitive` was used with only a single primitive type as the source. This not only increases the public interface much more than necessary but is also misleading. All fields that can be converted as such have specified widths iirc.

The `num-derive` crate does also not handle non-exhaustiveness which is a major future compatibility hazard as more tags etc. are to be supported.

Closes: #57, #59

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -10,27 +10,35 @@ use {TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
 use self::Value::{Ascii, List, Rational, Unsigned, Signed, SRational};
 
 macro_rules! tags {
-    {$(
-        $tag:ident
-        $val:expr;
-    )*} => {
-
-        /// TIFF tag
-        #[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
-        pub enum Tag {
-            $($tag,)*
-            Unknown(u16)
+    {
+        // Permit arbitrary meta items, which include documentation.
+        $( #[$enum_attr:meta] )*
+        pub enum $name:ident($ty:ty) {
+            // Each of the `Name = Val,` permitting documentation.
+            $($(#[$ident_attr:meta])* $tag:ident = $val:expr,)*
         }
-        impl Tag {
-            pub fn from_u16(n: u16) -> Tag {
-                $(if n == $val { Tag::$tag } else)* {
-                    Tag::Unknown(n)
+    } => {
+        $( #[$enum_attr] )*
+        #[derive(Clone, Copy, PartialEq, Eq, Debug, Hash)]
+        pub enum $name {
+            $($(#[$ident_attr])* $tag,)*
+            Unknown($ty),
+            #[doc(hidden)]
+            __NonExhaustive,
+        }
+
+        impl $name {
+            fn __from_inner_type(n: u16) -> Self {
+                $(if n == $val { $name::$tag } else)* {
+                    $name::Unknown(n)
                 }
             }
-            pub fn to_u16(&self) -> u16 {
+
+            fn __to_inner_type(&self) -> u16 {
                 match *self {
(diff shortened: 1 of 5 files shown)


src/decoder/ifd.rs
```diff
@@ -10,27 +10,41 @@ use {TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
...
+                #[doc = $unknown_doc]
+                Unknown($ty),
+            )*
         }
-        impl Tag {
-            pub fn from_u16(n: u16) -> Tag {
-                $(if n == $val { Tag::$tag } else)* {
-                    Tag::Unknown(n)
+
+        impl $name {
```
fintelia: I'm coming to this somewhat late so perhaps I'm missing some context, but is there a reason not to replace the methods below with impls of `TryFrom<$ty> for $tag` and `From<$tag> for $ty`?

HeroicKatora: I thought I'd minimize the public interface that comes from *within* the macro. It might still be a good idea to expose those.

HeroicKatora: We'll get `#[non_exhaustive]` (see https://github.com/rust-lang/rust/issues/44109) in a few releases and when we upgrade the compile requirements then we can definitely think of reworking this to purely perform discriminant casting as appropriate.

HeroicKatora:
> I generally don't like using macros if there are any reasonable alternatives, but I really don't see any here.

Yeah, sadly. We'll get `#[non_exhaustive]` (see https://github.com/rust-lang/rust/issues/44109) in a few releases and when we upgrade the compile requirements then we can definitely think of reworking this to purely perform discriminant casting but that still will not provide a safe option for `fn from_u16(u16) -> Option<Self>`... So yep

src/decoder/mod.rs
```diff
@@ -334,8 +379,9 @@ impl<R: Read + Seek> Decoder<R> {
         self.width = self.get_tag_u32(ifd::Tag::ImageWidth)?;
         self.height = self.get_tag_u32(ifd::Tag::ImageLength)?;
         self.strip_decoder = None;
-        self.photometric_interpretation = match FromPrimitive::from_u32(
-            self.get_tag_u32(ifd::Tag::PhotometricInterpretation)?
+        // TODO: error on non-SHORT value.
```
birktj: Shouldn't we just handle these errors right now?

HeroicKatora: It's not straightforward since the `Value` does not store whether it was create from a `SHORT` (`u16`) or a `BYTE` (`u8`) or another compatible type. The specification only permits exactly the `SHORT` type so that this is strictly more permissive either. And for any conforming image the `as u16` cast will also not result in incorrectly interpreted values.

Since it would entail a fair bit of rewrite of the whole `Value` system I'd rather do this in a future PR. I'll open a tracking issue.

birktj: Understandable

src/decoder/mod.rs
```diff
@@ -1,4 +1,4 @@
-use num_traits::{FromPrimitive, Num};
+use num_traits::{Num};
```
birktj: Removing the `num_traits` dependency as well should be pretty simple, right? Should only be a simple change in https://github.com/image-rs/image-tiff/blob/master/src/decoder/mod.rs#L194.

HeroicKatora: Huh, right. That function is only used twice.

birktj: To me it doesn't even seem like it uses the `Num` bound.

HeroicKatora: Right, I thought there was arithmetic going that includes the number type but the `Wrapping` trait has already been extracted.

Issue #48, opened 2019-07-27
Add deflate support


jmetz:
Presumably using e.g. [flate2](https://github.com/alexcrichton/flate2-rs) or similar should enable `deflate` support. 
The `CompressionMethod` should come out as `32946`.

Masterchef365:
See https://github.com/image-rs/image-tiff/pull/49!

Issue #60, opened 2019-10-11
Release a new version


nox:
It would be nice of you to release a new version with num-derive 0.3 so that downstream users can gradually get rid of syn 0.15 in their crate graph.

JohnTitor:
cc: @HeroicKatora, could you make a new release?

HeroicKatora:
There has been commit #3b85b68ae957e15e1e0609ea5368e849fdde9172 which is strictly speaking SemVer breaking. It adds a bunch of enum variants to `pub` types. From their use and definition it may have been somewhat obvious that these are intended to be non-exhaustive—they list a set of known, standardized identifiers—but the documentation didn't say so (in fact there is none) and they are a potential hazard. Is that an acceptable risk?

* Yes, release anyways :+1:
* No, revert the commit, do the release, re-apply the commit :-1:

Some input via respective reactions is appreciated.

nox:
There is no acceptable risk here, please make a breaking bump. Why would you want to revert anything? Both `tiff` and `image` can just do a breaking bump.

nox:
While at it, you could also just move the `ifd` stuff to its own crate, and make it an implementation detail in `tiff`, so that additional enum variants in any of the low-level blocks of `ifd` don't lead to a breaking change in the `tiff` crate.

HeroicKatora:
> Why would you want to revert anything? Both tiff and image can just do a breaking bump.

I think I slightly misunderstood the motivation, I thought it was very important that this were a *non SemVer breaking* release. Then bumping the major version of `tiff` is certainly the better solution, after explicitely marking the `enum` as non-exhaustive and addressing #59 in the process.

However, it is somewhat more difficult to 'just do a breaking bump of `image`' and it feels off since it doesn't even use any of the internal affected. There's still a large portion that uses *previous* major releases and don't get performance improvements, bugfix and *soundness* fixes. Hence, I must object to a breaking release in that crate for the sole reason of improving the dependency graph in one of the decoders. I'd be slightly more open to it if it were the last occurrance.

So, considering the above some engineering is required. Since the breaking changes to `tiff` did not do any bug fixes it seems most useful to make a `0.3.2` release without them, that's what the above comment tries to outline with reverting, releasing, re-applying.

HeroicKatora:
@nox What is the best course of action boils down to: Are you fine with waiting for `image = 0.23` until the changes are in or not?

nox:
This is no pressing matter, take your time, no worries.

[...]

PR #64, opened 2019-11-20
Document changes for release 0.4

HeroicKatora:
When this PR is merged then version 0.4 will be released. This serves as a last comment period since the crate is smaller and thus produces less immediate feedback.

Fixes: #48, #59, #60

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,12 @@
+# Version 0.4.0
+
+* Several enumerations are now non_exhaustive for future extensions.
+  These are `Tag`, `Type`, `Value`, `PhotometricInterpretation`,
+  `CompressionMethod`, `Predictor`.
+* Enums gained a dedicated method to convert to their TIFF variant value with
+  the specified type. Performing these conversions by casting the discriminant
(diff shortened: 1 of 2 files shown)


HeroicKatora:
Blocked on: #65 I'd say

CHANGES.md
```diff
@@ -0,0 +1,11 @@
+# Version 0.4.0
+
+* Several enumerations are now non_exhaustive for future extensions.
+  These are `Tag`, `Type`, `Value`, `PhotometricInterpretation`,
+  `CompressionMethod`, `Predictor`.
+* Enums gained a dedicated method to convert to their TIFF variant value with
+  the specified type. Performing these conversions by casting the discriminant
+  with `as` is not guaranteed to be stable, except where documented explicitly.
+* Removed the num-derive and num dependencies.
+* Added support for compressed images.
```
Robzz: Something like "Added support for decoding images compressed with the deflate method" would be more accurate here I think: we already supported some compression methods before that, at least for decoding (LZW and PackBits it looks like), what was added is deflate support in #49, and only for decoding apparently.

PR #65, opened 2019-11-20
Improve the error types

HeroicKatora:
This is a small addition intended to make it easier to complete unimplemented items with fewer breaking changes, similar to allowing the tags themselves to be expanded. It also expands the documentation of existing variants and exports the opaque encapsulation created for decompression which was silently hidden in a private module.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -26,7 +26,7 @@ pub enum TiffError {
 
 /// The image is not formatted properly.
 ///
-/// The indicate that the encoder producing the image might behave incorrectly or that the input
+/// This indicates that the encoder producing the image might behave incorrectly or that the input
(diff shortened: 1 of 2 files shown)


src/error.rs
```diff
@@ -11,19 +11,26 @@ use ColorType;
...
     IoError(io::Error),
 
-    /// The Limits of the Decoder is exceeded,
+    /// The Limits of the Decoder is exceeded.
     LimitsExceeded,
 }
 
+/// The image is not formatted properly.
+///
+/// The indicate that the encoder producing the image might behave incorrectly or that the input
```
Robzz: Typo here: should be "this indicates" I suppose.

PR #66, opened 2019-12-10
Support 32- and 64-bit integer sample grayscale and RGB(A) images

### LINKED ISSUE #19: Feature request: 32 bits per sample
Would you be able to add support for 32bit float Grayscale Tiffs?

aschampion:
(This issue originally mentioned #19, but that is for floating point images.)

Matching the existing style, there was no attempt to DRY code here, even if some of this could be generic'd or macro'd away. There are a finite number of types, after all. But if we wanted to support CMYK(>8), etc., or floating point images that may need to be done.

Though a BigTIFF tag data type for 64 bit uints (`LONG8`) was added to satisfy some TiffValue bounds for 64-bit encoding, this data type never has to be written/read for 64-bit image data. `libtiff` and most libraries understand this tag data type value anyway.


src/decoder/mod.rs
```diff
@@ -314,6 +364,18 @@ impl<R: Read + Seek> Decoder<R> {
...
+            }
+            PhotometricInterpretation::RGB if self.bits_per_sample == [32, 32, 32] => {
+                Ok(ColorType::RGB(32))
+            }
+            PhotometricInterpretation::RGB if self.bits_per_sample == [64, 64, 64, 64] => {
+                Ok(ColorType::RGBA(64))
+            }
+            PhotometricInterpretation::RGB if self.bits_per_sample == [64, 64, 64] => {
+                Ok(ColorType::RGB(64))
+            }
```
fintelia: This whole match could probably be more concise with a double nested match. Something along the lines of:
```rust
match self.photometric_interpretation {
     PhotometricInterpretation::RGB => match self.bits_per_sample {
         [8, 8, 8, 8] => Ok(ColorType::RGBA(8)),
         [16, 16, 16, 16] => Ok(ColorType::RGBA(16)),
         ...
     },
    ...
}

Comment (aschampion) [reply src/decoder/mod.rs]: Done in two commits, one making this simplification, and a second making a further simplification while also changing the behavior to better match `Gray` behavior, in accepting non-8, 16, 32 and 64 bit values so long as they are homogenous. This is preferred because `colortype` is just parsing the color type representation, not deciding if the decoder can read it.

There are some other small cleanups I'm putting in another PR that I want to land after this but before a PR that adds floating type support (which is already working).

Comment (fintelia) [reply src/decoder/mod.rs]: Are there uses in the wild of bit depths other than {8, 16, 32, 64}? We are in the process of removing support for strange bit depths in the main image crate's ColorType enum (master has them next has only common ones). If necessary, I'd prefer if this crate could "normalize" bit depths to ones that are reasonable to work with rather than pushing that responsibility off onto the user.

Comment (aschampion) [reply src/decoder/mod.rs]: Most notably 1, 2 and 4 bit images exist and people in scientific imaging have to deal with them. libtiff (C), imagej/bioformats (Java), and pillow and tifffile (python) all support them to varying extents. I'm fine with the main `image` crate converging on simple, common types, but the point of format-specific crates should be that they can support what the format supports, and the format module glue code in `image` can just accept the subset it understands. (Or eventually with `image-core` some of the glue code could be moved into the format crates themselves if that's preferred.)

I think the format specific crates should try to accept as much of what the format supports as possible (or at least be designed such that they don't rule out eventually supporting that), as that is a user's only reason to work with the format crate directly rather than the `image` crate. TIFF in particular is a very versatile spec whose oddities are often exploited.

Comment (aschampion) [reply src/decoder/mod.rs]: Whenever I have a few moments I'll add a test with some weird bit depth images to make sure they aren't decodable, but err in an ergonomic way and correctly indicate why they aren't yet supported.

Comment (aschampion) [reply src/decoder/mod.rs]: Tests added. I used 12-bit images as these are more common for both grayscale and RGB (whereas 1|2|4 bit are usually grayscale only) and also sometimes occur outside of scientific imaging in photography and industrial imaging. To also note with this change to the RGB matching, both the 12-bit grayscale and RGB image now error in the same way, whereas before they would have erred at different points.

File: src/decoder/mod.rs [hunk 2]
```diff
@@ -302,18 +352,18 @@ impl<R: Read + Seek> Decoder<R> {
     pub fn colortype(&mut self) -> TiffResult<ColorType> {
         match self.photometric_interpretation {
             // TODO: catch also [ 8, 8, 8, _] this does not work due to a bug in rust atm
```
fintelia: This TODO can be removed now right?

aschampion: Wasn't exactly sure, almost sounds like the original author wanted to match formats with different alpha channel depth, but the line history is difficult to follow back into the image crate. Since it isn't clear what the TODO would be besides "support inhomogeneous bit depths" (which would be a big, crate-wide undertaking and not something local to this statement), I'll remove it.

src/decoder/mod.rs
fintelia: The design I've been pushing towards for the main crate is to have decoders accept all valid images files, but have them to apply any necessary conversions so the resulting output is one of the ColorType variants. Then the [original_color_type](https://docs.rs/image/0.23.0-preview.0/image/trait.ImageDecoder.html#method.original_color_type) method serves to indicate what the color type was before any conversions. Encoders could do the same process in reverse, taking both a ColorType describing the input bytes as well as an ExtendedColorType describing the desired output pixel format. At some point we'd probably then provide a way to skip the extra decoding stage if desired.

(My position is that "decoding" should include not just transforming a file into an in memory grid of pixels, but also making sure that the pixel representation is something reasonable to work with.)

fintelia: In your 12-bit RGB example, the original color type would be Rgb12 while the output color type could be Rgb16. In this case the round-trip conversion between the two formats would be completely lossless, and there would be only a minor memory overhead.

aschampion: > In your 12-bit RGB example, the original color type would be Rgb12 while the output color type could be Rgb16.

This seems reasonable, so `colortype` is now doing the correct thing here in decoding the original color type (or our nearest representable match), while the output color type can be determined elsewhere in the decoder (and currently effectively is by the buffer selection in `new`, which I'm cleaning up in a follow up PR to prep for floating point support).

aschampion:
I think all the comments have been addressed. I can rename `Decoder::colortype` to better match the `original_color_type` discussion above, but would prefer to do that in the follow up refactoring PR to avoid having to resolve a bunch of merge conflicts.

fintelia:
Sounds good. Lets hold off on the renaming. Once image-core is split off and Decoder implements `image_core::ImageDecoder` we'll get consistent naming across everything. I'm not sure it is worth doing a breaking change before then

aschampion:
Is there a merge owner for this crate or should I merge myself?

aschampion:
Good to know, thanks!

PR #67, opened 2019-12-21
Refactoring, cleanup, safety, docs & CMYK(16|32|64) support

aschampion:
This is a collection of many mostly small changes that would be tedious to separate. Motivations:
- Preparing for floating point support, which is working and will be PR'd following this, but is much simpler with these changes than without.
- Preparing for BigTIFF support, which is still in progress but depends on these changes.
- Removing most unchecked casts, both for the purposes of robustness/correctness and as necessary preparation for generic bounds for BigTIFF.
- Improving some code structure and encoder/decoder code reuse.

Summary of major and notable breaking changes:
- Addition of a `TiffError::IntSizeError` top level variant to indicate size casting errors that don't necessarily fall into an existing error (i.e., may not be format errors or limit errors).
- Moving tags into a top-level module so they can be reused in the encoder/decoder and by users of the library. Some traits that may make it easier for consumers to use their own tag enums are also exposed.
- The README now summarizes supported and unsupported features, so users don't have to experiment to figure out what works.
- Add support for decoding and encoding CMYK(16|32|64). This is trivial due to the other refactoring and is mostly bringing features up to a logical symmetry rather than a new feature for a separate PR.

**A note about breaking changes:** only two published crates depend on `image-tiff`: `image`, which we control, and `n5`, which I control. A github code search found only one unpublished crate that did, [`ndimage`](https://github.com/Robzz/ndimage). Additionally `image-tiff` is pre-1.0 and still has a ways to go in feature support, so I'm in favor of being aggressive with breaking changes if they bring more structure necessary for full support of the format's capabilities.

PR is draft for the moment as I might amend a few more changes from my BigTIFF prototype into this.


HeroicKatora:
>  Additionally image-tiff is pre-1.0 and still has a ways to go in feature support, so I'm in favor of being aggressive with breaking changes if they bring more structure necessary for full support of the format's capabilities.

Supporting a whole new spectrum of color type, and refactoring encoding and decoding paths is an effort easily worth another major release. :+1:

> This is a collection of many mostly small changes that would be tedious to separate.

If it's possible, I could likely very quickly review the removal of `tempfile` and the related refactoring  and deduplication of tests. That's quite a bit of work that I'd wanted to do at some point but never quite found the time.

src/encoder/mod.rs
```diff
@@ -451,7 +445,7 @@ pub struct DirectoryEncoder<'a, W: 'a + Write + Seek> {
 impl<'a, W: 'a + Write + Seek> DirectoryEncoder<'a, W> {
     fn new(writer: &'a mut TiffWriter<W>) -> TiffResult<DirectoryEncoder<'a, W>> {
         // the previous word is the IFD offset position
-        let ifd_pointer_pos = writer.offset() - mem::size_of::<u32>() as u64;
+        let ifd_pointer_pos = writer.offset() - u64::try_from(mem::size_of::<u32>())?;
```
HeroicKatora: The `u64::try_from(_)` is a constant expression. I doesn't make a lot of sense to error at runtime here. I'd like to diagnose this at compile time but am not sure if it is pretty.

src/encoder/mod.rs
```diff
@@ -584,27 +580,27 @@ impl<'a, W: 'a + Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
         width: u32,
         height: u32,
     ) -> TiffResult<ImageEncoder<'a, W, T>> {
-        let row_samples = u64::from(width) * <T>::BITS_PER_SAMPLE.len() as u64;
+        let row_samples = u64::from(width) * u64::try_from(<T>::BITS_PER_SAMPLE.len())?;
```
HeroicKatora: Note for review readers: Reasoning would not be the same here, since `T` is an input type parameter. Panicking or build failures are not exactly welcome so keep the fallible conversion, imo.

src/error.rs
```diff
@@ -38,8 +42,9 @@ pub enum TiffFormatError {
     ImageFileDirectoryNotFound,
     InconsistentSizesEncountered,
     InvalidTag,
+    InvalidTagValueType(Tag),
     RequiredTagNotFound(Tag),
-    UnknownPredictor(u32),
+    UnknownPredictor(u16),
```
HeroicKatora: Good find!

src/tags.rs
```diff
@@ -0,0 +1,188 @@
+mod seals {
+    pub trait TagConversion<T>: Sized {
+        fn __from_inner_type(T) -> Result<Self, T>;
+
+        fn __to_inner_type(&self) -> T;
+    }
+}
+
+pub trait U16Tag: Sized {
```
HeroicKatora: I'm generally critical towards having *public* traits represent common inherent methods. Firstly, it's an evolution risk since they can be named in trait bounds and implemented by library users. Secondly, actually using the methods offered by them requires importing an additional `use` statement of an item which is only remotely connected. Thirdly, the methods are by default collapsed in the documentation and much harder to find (appear in the bottom, would expect mostly standard conversion functions, ..).

In essense, if the trait is not used in a cleaner interface I would rather expose the functionality as plain inherent methods. Feel free to do this via local macro rules, the standard library does it that way and this avoids duplication of code written.

src/encoder/mod.rs
aschampion: Yeah, don't know what I was thinking with this one -- it should just be `as u64` since `4` will obviously cast fine to `u64`, as would any real world `size_of`. This might be related to some generics I was doing for BigTiff branch, but if so I should change it there.

Eventually `unwrap` will be a const fn, which would allow a simple unwrap on `try_from` to throw at compile time or otherwise be optimized out in cases like this.

src/tags.rs
aschampion: These are inherent macro-generated methods now.

aschampion:
Haven't had much time for this lately (the curse of having a branch that does what you need is the lack of pressure to finish a PR), but the comments should be addressed now. I'd like to get this and the follow up float support PR through, but it will probably be some time before I can wrap up BigTiff.

HeroicKatora:
Thank you, so much again!

PR #70, opened 2020-02-06
Floating point sample support

aschampion:
Opportunities here for cleanup and deduplication, but those can come later.


src/error.rs
```diff
@@ -148,6 +149,9 @@ impl fmt::Display for TiffUnsupportedError {
             UnsupportedSampleDepth(samples) => {
                 write!(fmt, "{} samples per pixel is unsupported.", samples)
             }
+            UnsupportedSampleFormat(ref formats) => {
+                write!(fmt, "Sample format {:?} is unsupported.", formats)
```
HeroicKatora: Given that one (however common) possible error is that the formats are *mixed*, could we make an effort of trying to display these formats as decoded tags here?

src/decoder/mod.rs
```diff
@@ -675,6 +751,26 @@ impl<R: Read + Seek> Decoder<R> {
                 }
                 bytes
             }
+            (ColorType::Gray(32), DecodingBuffer::F32(ref mut buffer)) => {
+                reader.read_f32_into(&mut buffer[..bytes / 4])?;
+                if self.photometric_interpretation == PhotometricInterpretation::WhiteIsZero {
+                    for datum in buffer[..bytes / 4].iter_mut() {
+                        // FIXME: assumes [0, 1) range for floats
```
HeroicKatora: That seems fine, I'm remarking this in order to remind us of opening a tracking issue to investigate how this could be validated or how a violation of this assumption might be detected.

src/decoder/mod.rs
```diff
@@ -370,6 +422,18 @@ impl<R: Read + Seek> Decoder<R> {
...
             self.samples = val;
         }
+        if let Some(vals) = self.find_tag_unsigned_vec(Tag::SampleFormat)? {
+            self.sample_format = vals.iter()
+                .map(|&v| SampleFormat::from_u16(v)
+                    .ok_or_else(|| TiffUnsupportedError::UnsupportedSampleFormat(vals.clone()))
+                )
+                .collect::<Result<_, _>>()?;
+
+            // FIXME: for now, only homogenous formats across samples are supported.
```
HeroicKatora: That's more of a `TODO` than a `FIXME`.

src/decoder/mod.rs
```diff
@@ -197,6 +230,18 @@ impl Wrapping for u64 {
...
+    fn wrapping_add(&self, other: Self) -> Self {
+        self + other
+    }
+}
+
+impl Wrapping for f64 {
+    fn wrapping_add(&self, other: Self) -> Self {
+        self + other
+    }
+}
```
HeroicKatora: That's quite unusual, the assumption of predictors being that it is reversible. However, `f32` addition and subtraction is in fact lossy. Is this actually used in images?

src/decoder/mod.rs
aschampion: This is a bit complicated -- my understanding is pre-6.0 TIFF specs used `MinSampleValue` and `MaxSampleValue` as the limits that `WhiteIsZero` and `BlackIsZero` rescaled values into. By spec 6 this was dropped, and those fields became essentially meaningless. Separate `SMinSampleValue`/`SMaxSampleValue` were introduced, but these were just memoized ranges of the image and don't say anything about the sample value domain. The 6 spec just says 2^BitsPerSample, but that's because float is very much an afterthought in the spec. In practice, [0, 1) seems to be what most programs expect for float; we could check the `SMinSampleValue`/`SMaxSampleValue` (if present) to be inside this interval and error in this case.

src/tags.rs
```diff
@@ -180,4 +183,13 @@ pub enum ResolutionUnit(u16) {
     Inch = 2,
     Centimeter = 3,
 }
+}
+
+tags! {
+pub enum SampleFormat(u16) {
```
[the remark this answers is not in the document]
aschampion: All the complex image TIFFs I've ever encountered just encode the imaginary component as an extra sample (or even extra image) rather than using this type. I also can't find reference to this outside of the mailing list thread from 2000 when it was added to libtiff. So far my policy with tag values has been to only include ones we can interpret. If we want to list values we can't interpret but know exist, we can change that policy. I can see the motivation if one wanted to make an `identify` or `tiffinfo` like tool using this library.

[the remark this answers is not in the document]
aschampion: Also should be clear it's not in the spec, not that "being in a 30-year-old spec" is a good criteria for what to include and what not to with a flexible format like TIFF.

src/decoder/mod.rs
HeroicKatora: That's pretty much all of the information that I was missing here. If the missing parts are mostly checking for deprecated attributes and consistency checks that are seldomly relevant then this seems fine.

src/decoder/mod.rs
aschampion: I'll have to look into this -- there is a third predictor for floating point I wasn't aware of. However libtiff also seems to accept horizontal predictor for float types (at least the horizontal predictor code doesn't check the format type) by just casting the bytes to uints. Haven't dug through the spec yet, but it's enough to block this PR for now.

src/error.rs
aschampion: The error contains the decoded `SampleFormat` now (including unknown variant), so this should contain all the relevant information.

src/decoder/mod.rs
HeroicKatora: Found a specification, I think: http://chriscox.org/TIFFTN3d1.pdf

Farkal: Is this the only missing feature ? 
From the doc you link the `DecodeDeltaBytes` is the `rev_hpredict_nsamp` and the `DecodeFPDelta` could be implemented just by adding parameters to `rev_hpredict` and add some code to the current match. I am missing something ? 
Do we have some test files for this case ? All of my tiff are float with horizontal predictor.

Farkal:
I added the decoding of float tag (like MODELPIXELSCALE or MODELTRANSFORMATIONTAG) :wink: 
To do this i had to remove the impl of Eq and Hash on the ifd::Value and also on the error::TiffFormatError (because it take a ifd::Value as parameter). Float doesn't implement Eq and Hash because of the Nan issue. Is this required ? (the lib compile successfully).

src/decoder/mod.rs
HeroicKatora: It certainly is not that easy. I wrote the comment to refer to PREDICTOR_FLOATINGPOINT. The standard horizontal predictor should work _on byte data_ and not on floats and that is not implemented correctly here either. Your file should be a fine test case for these.

HeroicKatora:
> Float doesn't implement Eq and Hash because of the Nan issue. Is this required ? (the lib compile successfully).

I wouldn't say it's required. It might be somewhat annoying but not a deal breaker for me. `std::error::Error` also doesn't require `Eq` nor `Hash` so in the abstract common cases it shouldn't matter greatly for the error. And `PartialEq` still remains.

src/decoder/mod.rs
HeroicKatora: Also, I'm not quite sure where it's standardized. I'm somewhat grasping here because it integer expands samples for 4bit values and it would make sense to do something similar for floating point, i.e. reinterpret them. But I really can't find an exact reference for this.

HeroicKatora: Supporting this theory that floats are reinterpreted as bytes is [this python tiff library](https://github.com/cgohlke/tifffile/blob/987083705d1d988da2f27a3fb40784963cb2dab9/tifffile/tifffile.py#L12115-L12133).

src/decoder/mod.rs
```diff
@@ -238,6 +283,12 @@ fn rev_hpredict(image: DecodingBuffer, size: (u32, u32), color_type: ColorType)
         DecodingBuffer::U64(buf) => {
             rev_hpredict_nsamp(buf, size, samples)?;
         }
+        DecodingBuffer::F32(buf) => {
+            rev_hpredict_nsamp(buf, size, samples)?;
+        }
+        DecodingBuffer::F64(buf) => {
+            rev_hpredict_nsamp(buf, size, samples)?;
+        }
```
HeroicKatora, line 291: Better not to support those until we have right test data, I think. Having the data types and sample format checks is already a good part of the change.

Farkal: So you think we should just remove this and release floating point support ? And make another PR latter to add the support for the decoding buffer ?

HeroicKatora: Yes, not supporting the prediction seems less painful than a likely buggy implementation.

Farkal:
@HeroicKatora I am merging this branch with master but i have some troubles with the use of bytecast.
There is no function `from_le`and `from_be` for f32 and is it really more efficient to use the bytecast function to create `ne` bytes and then convert each one to `le` or `ne` ? (we could directly check the byteorder and convert to `le` or `ne`)

src/decoder/mod.rs
```diff
@@ -283,11 +283,13 @@ fn rev_hpredict(image: DecodingBuffer, size: (u32, u32), color_type: ColorType)
...
-            rev_hpredict_nsamp(buf, size, samples)?;
+        DecodingBuffer::F32(_buf) => {
+            unimplemented!()
+            // rev_hpredict_nsamp(buf, size, samples)?;
         }
-        DecodingBuffer::F64(buf) => {
-            rev_hpredict_nsamp(buf, size, samples)?;
+        DecodingBuffer::F64(_buf) => {
+            unimplemented!()
+            // rev_hpredict_nsamp(buf, size, samples)?;
```
HeroicKatora: Clarification: With 'not support this' I meant a new variant of [`TiffUnsupportedError`](https://docs.rs/tiff/0.5.0/tiff/enum.TiffUnsupportedError.html) not panicking.

Farkal:
Aw sorry I forgot this issue, is there something I can do to help the merge ?

PR #77, opened 2020-07-15
Read ascii encoded in offset

Farkal:
Currently the code try to read ascii encoded at the offset position but if it can be encoded on 4 bytes the tag is encoded in the offset and the code fail. I also fix an issue with the current read_string method.

I have tested this code by decoding geotiff file and in specialy the gdal_no_data tag of this image: https://github.com/georust/geotiff/blob/master/resources/zh_dem_25.tif

```rust
let img_file = File::open("resources/zh_dem_25.tif").expect("Cannot find test image!");
let mut decoder = Decoder::new(img_file).expect("Cannot create decoder");
println!("{:#?}", decoder);
let no_data = decoder.get_tag(tiff::tags::Tag::Unknown(42113)).expect("Failed to get tag");
println!("{:?}", no_data);
```

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -200,8 +200,8 @@ impl Entry {
                 }
                 if n <= 4 {
                     let mut buf = vec![0; n];
-                    self.r(bo).read_exact(&mut buf ).unwrap();
-                    let v = String::from_utf8(buf).unwrap();
+                    self.r(bo).read_exact(&mut buf )?;
+                    let v = String::from_utf8(buf)?;
                     let v = v.trim_matches(char::from(0));
                     Ok(Ascii(v.into()))
                 } else {
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -198,9 +198,17 @@ impl Entry {
                 if n > limits.decoding_buffer_size {
                     return Err(TiffError::LimitsExceeded);
                 }
-                decoder.goto_offset(self.r(bo).read_u32()?)?;
-                let string = decoder.read_string(n)?;
-                Ok(Ascii(string))
+                if n <= 4 {
+                    let mut buf = vec![0; n];
+                    self.r(bo).read_exact(&mut buf ).unwrap();
+                    let v = String::from_utf8(buf).unwrap();
+                    let v = v.trim_matches(char::from(0));
+                    Ok(Ascii(v.into()))
+                } else {
+                    decoder.goto_offset(self.r(bo).read_u32()?)?;
+                    let string = decoder.read_string(n)?;
+                    Ok(Ascii(string))
+                }
             }
             _ => Err(TiffError::UnsupportedError(
                 TiffUnsupportedError::UnsupportedDataType,
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -424,11 +424,13 @@ impl<R: Read + Seek> Decoder<R> {
     /// Reads a string
     #[inline]
     pub fn read_string(&mut self, length: usize) -> TiffResult<String> {
-        let mut out = String::with_capacity(length);
-        self.reader.read_to_string(&mut out)?;
+        let mut out = vec![0; length];
+        self.reader.read_exact(&mut out)?;
         // Strings may be null-terminated, so we trim anything downstream of the null byte
-        let trimmed = out.bytes().take_while(|&n| n != 0).collect::<Vec<u8>>();
-        Ok(String::from_utf8(trimmed)?)
+        if let Some(first) = out.iter().position(|&b| b == 0) {
+            out.truncate(first);
+        }
+        Ok(String::from_utf8(out)?)
     }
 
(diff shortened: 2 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -424,11 +424,13 @@ impl<R: Read + Seek> Decoder<R> {
...
-        self.reader.read_to_string(&mut out)?;
+        let mut out = vec![0; length];
+        self.reader.read_exact(&mut out)?;
         // Strings may be null-terminated, so we trim anything downstream of the null byte
-        let trimmed = out.bytes().take_while(|&n| n != 0).collect::<Vec<u8>>();
-        Ok(String::from_utf8(trimmed)?)
+        if let Some(first) = out.iter().position(|&b| b == 0) {
+            out.truncate(first);
+        }
+        Ok(String::from_utf8(out)?)
```
HeroicKatora, line 433: This change is because otherwise we'd require utf-8 encoding after the null terminator, right?
Do you have a small regression test? Would it be possible to create one?

src/decoder/ifd.rs
```diff
@@ -198,9 +198,17 @@ impl Entry {
                 if n > limits.decoding_buffer_size {
                     return Err(TiffError::LimitsExceeded);
                 }
-                decoder.goto_offset(self.r(bo).read_u32()?)?;
-                let string = decoder.read_string(n)?;
-                Ok(Ascii(string))
+                if n <= 4 {
+                    let mut buf = vec![0; n];
```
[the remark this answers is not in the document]
Farkal: Yeah you are right. But the function `String::from_utf8` take a vec so it will be on the heap anyway no ?
```rust
                if n <= 4 {
                    let mut buf = [0u8; 4];
                    self.r(bo).read_exact(&mut buf )?;
                    let v = String::from_utf8(buf.into())?;
                    let v = v.trim_matches(char::from(0));
                    Ok(Ascii(v.into()))
                } else
                ...
```

src/decoder/mod.rs
Farkal: It happen when i encode -9999 in the ASCII tag with gdal: `gdal_edit.py -a_nodata -9999 resources/zh_dem_25.tif`

The output bytes are `[0, 0, 0, 0, 0, 0]`, so i think creating a tag with the same value should create the bug we had with invalid utf-8 string.

src/decoder/ifd.rs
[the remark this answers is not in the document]
HeroicKatora: Stupid me (*facepalm*). Of course, the String will reuse of the allocation in the common case. Then the only advantage would be that the allocation might be moved _after_ the trimming of null bytes, so a string of only nulls would be come empty and wouldn't require one. I'm not sure if that is common enough (or even valid?) to warrant the special case..

src/decoder/mod.rs
HeroicKatora: The byte `\0` is a valid utf-8 character though, so that should work with the new and old code alike.

src/decoder/ifd.rs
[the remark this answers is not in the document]
Farkal: I can push the code, it will be optimized for the case you mention anyway ;)

src/decoder/mod.rs
Farkal: Yeah, i can keep the same code and just change the out into iter:
```rust
// Strings may be null-terminated, so we trim anything downstream of the null byte
let trimmed = out.into_iter().take_while(|&n| n != 0).collect::<Vec<u8>>();
```
The main issue was the `read_to_string` function.

HeroicKatora: I don't understand? `truncate` is surely more performant than collecting individual bytes in a new vector. No, this current code is definitely more correct. I was only unsure how this issue was found in the first case since `read_to_string` should have no problem with reading `[0, 0, 0, 0, 0, 0]`.

Farkal: Excuse me the value is `[45, 57, 57, 57, 57, 0]`
Ok i thought your impl was more performant than mine.

HeroicKatora: Are there by any chance any bytes after the null terminator?

HeroicKatora: I'm still not sure why this would fail with the previous implementation otherwise.

Farkal: I also don't understand why this fail with the method read_to_string
Here is more bytes:
`[45, 57, 57, 57, 57, 0, 0, 0, 0, 0, 0, 0, 57, 64, 0, 0, 0, 0, 0, 0, 57, 64, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]`

HeroicKatora: Yeah, if any of the values were larger than `128` then it might no longer be valid utf8. Such a tail would be rejected by `read_to_string` but accepted by the new variant, which seems more correct. In any case I was only trying to understand if it were easy to create a reduced regression test (<4kB) for this. It's not critical for the PR.

PR #79, opened 2020-07-15
Export tag macro rule

Farkal:
I don't know how we are supposed to implement our own tags but this is how i did:

```rust
use tiff::tags;
use tiff::tags::Tag;

tags! {
pub enum GeoTiffTag(u16) {
    // GDAL
    GdalMetadata = 42112,
    GdalNodata = 42113,

    // GeoTiff
    ModelPixelScale = 33550,
    ModelTiepoint = 33922,
    ModelTransformationTag = 34264,
    GeoKeyDirectoryTag =34735,
    GeoDoubleParamsTag= 34736,
    GeoAsciiParamsTag =34737,
}
}

impl From<GeoTiffTag> for Tag {
    fn from(g: GeoTiffTag) -> Self {
        Tag::Unknown(g.to_u16())
    }
}
```

Then i can easily call the decoder with my own tags:
```rust
let no_data = decoder.get_tag(GeoTiffTag::GdalNodata.into()).expect("Failed to get tag");
```

Is there a better way to implement my own tags and use it with the decoder ?

### DIFF
--- a/src/tags.rs
+++ b/src/tags.rs
@@ -1,3 +1,4 @@
+#[macro_export]
 macro_rules! tags {
     {
         // Permit arbitrary meta items, which include documentation.


fintelia:
Couldn't you just do something like:
```rust
const GdalMetadata: Tag = Tag::Unknown(42112);
const GdalNodata: Tag = Tag::Unknown(42113);

let no_data = decoder.get_tag(GdalNodata).expect("Failed to get tag");
```
Am I missing something?

HeroicKatora:
The extra semantics of explicit variants is _something_. However, you could add a converting `From` impl to create the `Tag` but using the fitting value of `Unknown`. You do not have to use the `tags` macro for it. The macro exists since many definitions use the same mechanisms of non-exhaustiveness and `Unknown`-fallback variant. There are no special semantics attached to it. In particular since `GeoTiffTag` would only model the GeoTiff tags and does not have an unknown variant I don't think it fits even particular well for that use.

Farkal:
> Couldn't you just do something like:
> 
> ```rust
> const GdalMetadata: Tag = Tag::Unknown(42112);
> const GdalNodata: Tag = Tag::Unknown(42113);
> 
> let no_data = decoder.get_tag(GdalNodata).expect("Failed to get tag");
> ```
> 
> Am I missing something?

No you aren't, it work perfectly :smile: 

I should be usefull to also have an unknow tag in the geotiff lib (for the cloud optimized geotiff for exemple) but i think it's better to use the one of this lib. I can just wrap it and expose it if i need.

So we can close this merge request.

PR #81, opened 2020-07-21
Update link to TIFF specification in docs

spenczar:
Adobe refactored their site and the specification has moved.


HeroicKatora:
Thank you!

Issue #82, opened 2020-07-22
JPEG Compression Tag


powturns:
I noticed that `CompressionMethod::JPEG` has the value `6`. According to [wikipedia](https://en.wikipedia.org/wiki/TIFF#TIFF_Compression_Tag), this is actually the "obsolete 'old-style' JPEG" that "... should never be written". For consistency with `Deflate`/`OldDeflate` should we consider renaming this to `OldJPEG`?

I realize this is a breaking change, but since it seems like there is some work being done around this area it may be a good opportunity to standardize the names.

HeroicKatora:
I would just deprecate the old variant. It matches the intended semantics and is not a breaking change.

51systems:
If we incorporate the modern JPEG (7), would that be called `NewJPEG`?

HeroicKatora:
I prefer the implicit suggestion, `ModernJPEG` :grinning:

PR #95, opened 2020-09-25
Support decoding of modern JPEG compression

ruffson:
Implements decoding part of #82.
# Description

Hello there,

I wrote some code to make this package compatible to the "modern JPEG" as described in the famous [technote-2](https://www.awaresystems.be/imaging/tiff/specification/TIFFTechNote2.txt).

This is my first attempt at contributing to a Rust package so please keep this in mind. As the technote states, the whole point of this revamped JPEG support was to be able to use available decoders. So this is why I used the `jpeg-decoder` package.

A few notes about this PR:
- According to `technote-2` (p.7), the`JPEGTABLES` tag can include either quantization tables, huffman tables or both. I tried with one alone and with both and it worked fine. I forgot to account for an empty `JPEGTABLES` (see open tasks below), but that should be easy enough. (Just a check)
- I successfully tested with RGB and grayscale JPEG-TIFFs and it worked fine. If others could test their images, please report problems and share your images, if possible
- In `decoder/mod.rs`: I did not follow the whole calling order of calling `read_image() -> read_strip_to_buffer() -> expand_strip()` but instead I created a new function `read_jpeg()` which is called from `read_image()`.
- I did not implement tile-wise decoding although I included the necessary tags and I want to give it a try for jpeg if it is easy enough to do.

# Open tasks
- [x] Make sure that this works with empty `JPEGTABLES`, e.g. account for the case where every segment includes its own quantization and huffman tables. (If someone can supply an example file, that would make this faster)
### Optional:
~~- [ ] Enable tile-wise reading (or remove the code that reads it and display a meaningful error message)~~ postponed to a later point in time, will make a new PR if I get this right.

---

I'll take a stab at tile-wise reading over the weekend, if this is not easily doable I'll drop it as I need to work with this soon. After this went through I want to contribute geoTIFF support (already working on that).

If people can test this, tell me what doesn't work and if someone could review my code, that would be very much appreciated. ~~Please note that some of the changes are due to auto-formatting.~~ Just rebased...

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -10,8 +10,8 @@ use tags::{Tag, Type};
 use {TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
 
 use self::Value::{
-    Ascii, Byte, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Signed, SignedBig,
-    Unsigned, UnsignedBig,
+    Ascii, Byte, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Signed,
+    SignedBig, Unsigned, UnsignedBig,
 };
 
 #[allow(unused_qualifications)]
@@ -38,9 +38,7 @@ impl Value {
     pub fn into_u8(self) -> TiffResult<u8> {
         match self {
             Byte(val) => Ok(val),
-            val => Err(TiffError::FormatError(
-                TiffFormatError::ByteExpected(val),
-            )),
+            val => Err(TiffError::FormatError(TiffFormatError::ByteExpected(val))),
         }
     }
 
@@ -452,7 +450,7 @@ impl Entry {
                         let string = decoder.read_string(n)?;
                         Ok(Ascii(string))
                     }
-                },
+                }
                 (Type::UNDEFINED, n) => self.decode_offset(n, bo, limits, decoder, |decoder| {
                     Ok(Byte(u8::from(decoder.read_byte()?)))
                 }),
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -425,7 +425,10 @@ impl Entry {
                         let string = decoder.read_string(n)?;
                         Ok(Ascii(string))
                     }
-                }
+                },
+                (Type::UNDEFINED, n) => self.decode_offset(n, bo, limits, decoder, |decoder| {
+                    Ok(Unsigned(u32::from(decoder.read_byte()?)))
+                }),
                 _ => Err(TiffError::UnsupportedError(
                     TiffUnsupportedError::UnsupportedDataType,
                 )),
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -10,13 +10,14 @@ use tags::{Tag, Type};
 use {TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
 
 use self::Value::{
-    Ascii, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Signed, SignedBig,
+    Ascii, Byte, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Signed, SignedBig,
     Unsigned, UnsignedBig,
 };
 
 #[allow(unused_qualifications)]
 #[derive(Debug, Clone, PartialEq)]
 pub enum Value {
+    Byte(u8),
     Signed(i32),
     SignedBig(i64),
     Unsigned(u32),
@@ -34,6 +35,16 @@ pub enum Value {
 }
 
 impl Value {
+
+    pub fn into_u8(self) -> TiffResult<u8> {
+        match self {
+            Byte(val) => Ok(val),
+            val => Err(TiffError::FormatError(
+                TiffFormatError::UnsignedIntegerExpected(val),
+            )),
+        }
+    }
+
     pub fn into_u32(self) -> TiffResult<u32> {
         match self {
             Unsigned(val) => Ok(val),
@@ -108,6 +119,23 @@ impl Value {
                 Ok(vec![u32::try_from(numerator)?, u32::try_from(denominator)?])
             }
             Ascii(val) => Ok(val.chars().map(u32::from).collect()),
+            val => Err(TiffError::FormatError(
+                TiffFormatError::UnsignedIntegerExpected(val),
+            )),
+        }
+    }
+
+    pub fn into_u8_vec(self) -> TiffResult<Vec<u8>> {
+        match self {
+            List(vec) => {
+                let mut new_vec = Vec::with_capacity(vec.len());
+                for v in vec {
+                    new_vec.push(v.into_u8()?)
+                }
+                Ok(new_vec)
+            }
+            Byte(val) => Ok(vec![val]),
+
             val => Err(TiffError::FormatError(
                 TiffFormatError::UnsignedIntegerExpected(val),
             )),
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -35,7 +35,6 @@ pub enum Value {
 }
 
 impl Value {
-
     pub fn into_u8(self) -> TiffResult<u8> {
         match self {
             Byte(val) => Ok(val),
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -39,7 +39,7 @@ impl Value {
         match self {
             Byte(val) => Ok(val),
             val => Err(TiffError::FormatError(
-                TiffFormatError::UnsignedIntegerExpected(val),
+                TiffFormatError::ByteExpected(val),
             )),
         }
     }
@@ -454,7 +454,7 @@ impl Entry {
                     }
                 },
                 (Type::UNDEFINED, n) => self.decode_offset(n, bo, limits, decoder, |decoder| {
-                    Ok(Unsigned(u32::from(decoder.read_byte()?)))
+                    Ok(Byte(u8::from(decoder.read_byte()?)))
                 }),
                 _ => Err(TiffError::UnsupportedError(
                     TiffUnsupportedError::UnsupportedDataType,
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -957,40 +957,25 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(())
     }
 
-    pub fn read_jpeg_tiles(&mut self) -> TiffResult<DecodingResult> {
-        let tile_offsets = self.get_tag_u32_vec(Tag::TileOffsets)?;
-        let tile_bytes = self.get_tag_u32_vec(Tag::TileByteCounts)?;
+    pub fn read_jpeg(&mut self, read_strip_wise: bool) -> TiffResult<DecodingResult> {
+
+        let offsets: Vec<u32>;
+        let bytes: Vec<u32>;
         let jpeg_tables: Vec<u8> = self.get_tag_u8_vec(Tag::JPEGTables)?;
 
-        let mut res_img = Vec::with_capacity(tile_offsets[0] as usize);
-
-        for (idx, offset) in tile_offsets.iter().enumerate() {
-            self.goto_offset(*offset)?;
-            let jpeg_reader = JpegReader::new(&mut self.reader, tile_bytes[idx], &jpeg_tables)?;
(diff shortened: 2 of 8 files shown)


src/decoder/mod.rs
```diff
@@ -710,6 +712,32 @@ impl<R: Read + Seek> Decoder<R> {
...
+                _ => {
+                    println!("Could not unwrap {:?}", key);
+                    continue;
+                }
+            };
+            all_tags.push((key, tag_value));
+        }
+
+        all_tags
+    }
```
HeroicKatora, line 739: This too appears to be on its own and unused. There's a lot of design space available for _just_ this method and I don't want to hold up the actual content of jpeg decoding over discussing this for too long.

src/decoder/ifd.rs
```diff
@@ -10,13 +10,14 @@ use tags::{Tag, Type};
...
 use self::Value::{
-    Ascii, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Signed, SignedBig,
+    Ascii, Byte, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Signed, SignedBig,
     Unsigned, UnsignedBig,
 };
 
 #[allow(unused_qualifications)]
 #[derive(Debug, Clone, PartialEq)]
 pub enum Value {
+    Byte(u8),
```
HeroicKatora, line 20: Note how `short` is implemented, using `Unsigned`. We will want to change this at some point but for now please use the same style.

src/decoder/mod.rs
```diff
@@ -923,6 +956,38 @@ impl<R: Read + Seek> Decoder<R> {
...
+
+        if read_strip_wise {
+            offsets = self.get_tag_u32_vec(Tag::StripOffsets)?;
+            bytes = self.get_tag_u32_vec(Tag::StripByteCounts)?;
+        } else {
+            offsets = self.get_tag_u32_vec(Tag::TileOffsets)?;
+            bytes = self.get_tag_u32_vec(Tag::TileByteCounts)?;
+        }
+
+        let mut res_img = Vec::with_capacity(offsets[0] as usize);
```
HeroicKatora, line 973: Allocations should be checked against the memory limits.

src/decoder/mod.rs
```diff
@@ -923,6 +956,38 @@ impl<R: Read + Seek> Decoder<R> {
...
+
+        for (idx, offset) in offsets.iter().enumerate() {
+            self.goto_offset(*offset)?;
+            let jpeg_reader = JpegReader::new(&mut self.reader, bytes[idx], &jpeg_tables)?;
+            let mut decoder = jpeg::Decoder::new(jpeg_reader);
+
+            match decoder.decode() {
+                Err(e) => {
+                    panic!("Error when reading jpeg tiles: {}", e);
```
HeroicKatora, line 982: Never panic on an error from input. Instead, match the underlying [`Error`](https://docs.rs/jpeg-decoder/0.1.20/jpeg_decoder/enum.Error.html) and return the io error, or wrap into a new variant of `TiffFormatError`/`UnsupportedError` as appropriate.

src/decoder/mod.rs
```diff
@@ -923,6 +956,38 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(())
     }
 
+    pub fn read_jpeg(&mut self, read_strip_wise: bool) -> TiffResult<DecodingResult> {
```
HeroicKatora, line 959: Is the caller ever supposed to know the right value of `read_strip_wise` through some other channel? If not, I'd suggest remove the parameter and moving the code reading the `TileOffset` tag to here.

src/decoder/mod.rs
```diff
@@ -1045,6 +1110,14 @@ impl<R: Read + Seek> Decoder<R> {
 
     /// Decodes the entire image and return it as a Vector
     pub fn read_image(&mut self) -> TiffResult<DecodingResult> {
+
+        if self.compression_method == CompressionMethod::ModernJPEG {
+            match self.find_tag(Tag::TileOffsets) {
+                Ok(None) | Err(_) => return self.read_jpeg(true),
```
HeroicKatora, line 1116: We could argue that ignoring some errors is fine for falling back but _all_ errors including `Io` is too general.

src/decoder/mod.rs
HeroicKatora: Also, we haven't validated that `offsets` is not empty. In this case, which is an invalid image, this code would panic but instead should return an error.

src/decoder/stream.rs
```diff
@@ -306,6 +318,72 @@ impl EndianReader for LZWReader {
...
 }
 
+///
+/// ## JPEG Reader (for "new-style" JPEG format (TIFF compression tag 7))
+///
+
+pub struct JpegReader {
+    buffer: io::Cursor<Vec<u8>>,
+    byte_order: ByteOrder, // for jpeg image data always ...,
+}
```
HeroicKatora, line 328: I'd rather we keep this `pub(crate)` for now. I'm not sure if we want to commit to supporting this exact interface for the long term, also due to some efficiency concerns.

src/decoder/mod.rs
```diff
@@ -923,6 +956,38 @@ impl<R: Read + Seek> Decoder<R> {
...
+        } else {
+            offsets = self.get_tag_u32_vec(Tag::TileOffsets)?;
+            bytes = self.get_tag_u32_vec(Tag::TileByteCounts)?;
+        }
+
+        let mut res_img = Vec::with_capacity(offsets[0] as usize);
+
+        for (idx, offset) in offsets.iter().enumerate() {
+            self.goto_offset(*offset)?;
+            let jpeg_reader = JpegReader::new(&mut self.reader, bytes[idx], &jpeg_tables)?;
```
HeroicKatora, line 977: You'll also want to validate this `idx`.

src/decoder/stream.rs
```diff
@@ -306,6 +318,72 @@ impl EndianReader for LZWReader {
...
+        jpeg_tables: &Vec<u8>,
+    ) -> io::Result<JpegReader>
+    where
+        R: Read + Seek,
+    {
+        let order = reader.byte_order;
+
+        // Read jpeg image data
+        let mut segment = vec![0; length as usize];
+        reader.read_exact(&mut segment[..])?;
```
HeroicKatora, line 355: Reading into an allocated buffer is okay for now, albeit somewhat wasteful. However, again, you should ensure the size of the allocation is limited.

HeroicKatora (changes requested):
I tried to give a quick review on the larger approach and style. Thanks for being detailed in your PR 
description. 


>  According to technote-2 [..]. I forgot to account for an empty JPEGTABLES (see open tasks below), but that should be easy enough. (Just a check)

Be sure to include that as a comment in the code.

> I successfully tested with RGB and grayscale JPEG-TIFFs and it worked fine. 

:tada: 

> In decoder/mod.rs: I did not follow the whole calling order of calling read_image() -> read_strip_to_buffer() -> expand_strip() but instead I created a new function read_jpeg() which is called from read_image().

That seems fine.

> I did not implement tile-wise decoding although I included the necessary tags and I want to give it a try for jpeg if it is easy enough to do.

This is also fine but try to explicitly error with a (new) variant of `Unsupported` in the cases where you already anticipate missing functionality. Nothing is harder to maintain or handle than incomplete interfaces that appear to work.

Don't be too surprised if I missed smaller issues, my priority was to avoid You investing too much work in detail before adjusting to the more general concerns.

ruffson:
Thank you so much, will tackle these things today!

src/decoder/mod.rs
ruffson: Yes this was useful for me. Do you want me to remove it?

src/decoder/ifd.rs
ruffson: I don't get this. Lets say I remove `Byte` from `Value` and change the part in the `val()` function to:

```rust
                (Type::UNDEFINED, n) => self.decode_offset(n, bo, limits, decoder, |decoder| {
                    offset_to_bytes(8, self)
                }),
```
I would still need the `into_u8()` and `into_u8_vec()` but with what type?

src/decoder/mod.rs
HeroicKatora: Yes please. Both printing and allocating the values is questionable for the public interface of the library. We can discuss this in a separate issue since I do see how it could be useful.

ruffson: Oh dang forgot the print. Will do.

src/decoder/stream.rs
ruffson: Okay, two questions:
1. what would be a less wasteful approach?
2. What do you mean with "ensuring that the allocation is limited?"

src/decoder/ifd.rs
HeroicKatora: Actually it's fine. I think we'll want to rework the internals of `Value` for the next major release so this shouldn't affect anything. Keep it the way it is :slightly_smiling_face:

ruffson: Okay!

src/decoder/stream.rs
HeroicKatora: > 1. what would be a less wasteful approach?

Wasteful refers to the fact that all data is copied into `segment`, then read into another buffer when the jpeg reader requests it–plus the clone of `jpeg_tables`. More efficient would be to avoid this intermediate buffering but it's somewhat complicated and definitely intermediate Rust. Very roughly:

* Keep the borrowed `&mut SmartReader<R>` in the struct.
* Also store the bytes of `jpeg_values` as a references (`&[u8]`).
* Complicate the impl of `Read` to first read from the values and when those are exhausted, forward to SmartReader.

But that gets tricky and its definitely not required to accept the PR.

> 2. What do you mean with "ensuring that the allocation is limited?"

The length is from `bytes[idx]` which is basically input data. We don't want to allocate 4GB of memory just because the image tells us to. You will find in the `Limits` struct some configuration for maximums. Basically you should check that the length does not exceed the `decoding_buffer_size`, although maybe even adding a third limit—named `intermediate_buffer_size` with default maybe ~128MB?—might be more accurate. This doesn't need to be 100% exact, i.e. sometimes we allocate more than once, but it should be close.

Note how the approach outlined in `1.` avoids having such a separate limit by not having an intermediate buffer.

ruffson: Thank you for the time to respond! 

> - Complicate the impl of Read to first read from the values and when those are exhausted, forward to SmartReader.

Ah this was my first idea, but I did not want to have too many moving parts, manage state and fight Rust. So I was trying to keep it simple and closely followed the approach taken in the `LZWReader`.

> ...Limits...

Ah okay I understand, will do. Although I find the size limitations overall pretty strict. E.g. I cannot load a measly 300MB TIF without having to manually adjust the limits. (But I get that limits have to be scaled up conservatively as different RAM configurations can cause problems and might require mmap)

ruffson:
Okay I tackled everything that came up so far. Please check if this still makes sense.

Then, what about the CI checks? I don't really understand why they are failing, are they testing against an older Rust version? I can run everything fine on my machine.

HeroicKatora:
>  I don't really understand why they are failing, are they testing against an older Rust version?

Yes, in particular against `1.34.2` with edition 2015. There have been some slightly relaxations to the borrow checker. In this case, you should be able to simply split the line into two to get around the check: First calculate the length and then truncate.

ruffson:
Cool that went through now.

src/decoder/mod.rs
```diff
@@ -923,6 +934,60 @@ impl<R: Read + Seek> Decoder<R> {
...
+        };
+
+        if offsets.len() == 0 {
+            return Err(TiffError::FormatError(TiffFormatError::RequiredTagEmpty(
+                Tag::StripOffsets,
+            )));
+        }
+        if offsets.len() != bytes.len() {
+            return Err(TiffError::FormatError(TiffFormatError::Format(
+                String::from("Size mismatch of StripOffsets and StripByteCounts"),
```
HeroicKatora, line 954: This unecessary allocates a string. The usual strip based decoding uses `TiffFormatError::InconsistentSizesEncountered` for this, please do so as well.

HeroicKatora (changes requested):
Looks pretty much done, there's just a small thing I noticed in one error case.

src/decoder/mod.rs
ruffson: done!

HeroicKatora (approved):

ruffson:
Awesome, thank you!

HeroicKatora:
Thanks!

PR #96, opened 2020-10-10
Add ScanImage BigTiff decoder

HagaiHargil:
Hi,

In my scientific field (neuroscientific imaging) we usually record tiffs in the [ScanImage BigTiff format](http://scanimage.vidriotechnologies.com/display/SI2016/ScanImage+BigTiff+Specification#ScanImageBigTiffSpecification-Non-VaryingFrameData), which is a format designed for microscopists. It's very similar to the standard BigTiff format, but it does add two blocks of metadata that can be parsed during the initial header parsing step.

This PR adds support for this data format, following the introduction of BigTiff support to this library. The PR is not complete due to a few issues:
1. I'm generally new to Rust, so this will definitely show.
2. I've added a special error type that reports failure to parse ScanImage's metadata, but I'm not sure how to make my code use it again due to my inexperience in Rust.
3. I haven't added a test that actually reads the entirety of the demo file I uploaded. This is because I wanted to first make sure I'm not doing something completely off in this PR. In principle, though, the standard BigTiff parsing code should work here as well.
4. The metadata is parsed as a string, but it's actually more like a hashmap (the ROI section) or a JSON-like document (the non-varying frame data section). I wasn't sure how to handle that currently, so I kept them as 'raw' strings.

Another possible blocking issue is that the demo file is a bit big (3 MB I think). I'll try to make a new one ASAP, depending on the status of this PR of course.

Thanks for this great library!


src/decoder/mod.rs
```diff
@@ -185,6 +185,28 @@ impl Default for Limits {
...
+/// a None if the file is not a ScanImage BigTiff.
+#[derive(Debug, Clone)]
+pub struct ScanImageMetadata {
+    /// Bytes 4 to 8 of the static metadata section
+    pub tif_version: u32,
+    /// Non-varying frame data
+    pub nonvar_frame_data: String,
+    /// ROI group data
+    pub roi_group_data: String,
+}
```
HeroicKatora, line 201: This should have a private member `_private: ()` to ensure that further fields can be added without a breaking change. I do not know enough about the format to judge if this is likely _but_ without an open standard I'd rather be cautious than require a major bump for adding/changing support for the metadata.

src/decoder/mod.rs
```diff
@@ -185,6 +185,28 @@ impl Default for Limits {
...
+
+
+impl ScanImageMetadata {
+    pub fn new(tif_version: u32, nonvar_frame_data: String, roi_group_data: String) -> Self {
+       ScanImageMetadata { tif_version, nonvar_frame_data, roi_group_data }
+    }
```
HeroicKatora, line 207: There doesn't need to be any public constructor, unless You want to add encoding support as well?

src/decoder/mod.rs
```diff
@@ -346,6 +370,10 @@ impl<R: Read + Seek> Decoder<R> {
     pub fn dimensions(&mut self) -> TiffResult<(u32, u32)> {
         Ok((self.width, self.height))
     }
+    
+    pub fn scanimage_metadata(&self) -> Option<ScanImageMetadata> {
+        self.scanimage.clone()
+    }
```
HeroicKatora, line 376:
```suggestion
    pub fn scanimage_metadata(&self) -> Option<&ScanImageMetadata> {
        self.scanimage.as_ref()
    }
```

src/decoder/mod.rs
```diff
@@ -429,6 +458,36 @@ impl<R: Read + Seek> Decoder<R> {
             0 => None,
             n => Some(n),
         };
+        // A special BigTiff format is generated by ScanImage and is identified
+        // and (partially) parsed under this if statement. It's identified by
+        // matching on a "magic number" that ScanImage defines [here](http://scanimage.vidriotechnologies.com/display/SI2019/ScanImage+BigTiff+Specification#ScanImageBigTiffSpecification-ScanImageStaticMetadataSection).
+        if self.bigtiff {
```
HeroicKatora, line 464: In general I would say that this should require an 'opt-in' of some sort. As far as I understand, the ifd offset is non-zero such that generic big tiff reading can simply skip over the extra metadata. But what's to say that some other image does not start with the magic number by chance anyways? Then this might error where it previously did not.

I'd propose adding a field `read_scan_image: bool` that starts as `false` and which you explicitly activate by calling a method similar to limits.

> `pub fn with_scan_image(mut self) -> Decoder<R> `

Then this check would become:
```suggestion
        if self.bigtiff && self.read_scan_image {
```

src/decoder/mod.rs
```diff
@@ -429,6 +458,36 @@ impl<R: Read + Seek> Decoder<R> {
...
+        let roi_group_data_length = self.read_long()?;
+        let mut nonvar_frame_data = vec![0u8; usize::try_from(nonvar_frame_data_length)?];
+        self.reader.read_exact(&mut nonvar_frame_data)?;
+        let mut roi_group_data = vec![0u8; usize::try_from(roi_group_data_length)?];
+        self.reader.read_exact(&mut roi_group_data)?;
+        // Parse the read bytes into UTF8 and into the metadat object
+        let nonvar_frame_data = String::from_utf8(nonvar_frame_data)?;
+        let roi_group_data = String::from_utf8(roi_group_data)?;
+        self.scanimage = Some(
+            ScanImageMetadata::new(tif_version, nonvar_frame_data, roi_group_data)
```
HeroicKatora, line 489: No need to have a dedicated constructor method. It's cleaner to use `ScanImageMetadata { }` and fill in the fields.

src/decoder/mod.rs
```diff
@@ -597,6 +656,20 @@ impl<R: Read + Seek> Decoder<R> {
...
     }
 
+    /// Returns `true` if the file is a BigTiff
+    pub fn is_bigtiff(&self) -> bool {
+        self.bigtiff
+    }
+
+    /// Returns the values located in the next n bytes of the file, where n is
+    /// the length of the given buffer
+    pub fn peek_n_bytes(&mut self, buffer: &mut [u8]) -> Result<(), io::Error> {
```
HeroicKatora, line 666: I don't think this was meant to be public?

src/decoder/mod.rs
```diff
@@ -429,6 +458,36 @@ impl<R: Read + Seek> Decoder<R> {
...
+        let tif_version = self.read_long()?;
+        let nonvar_frame_data_length = self.read_long()?;
+        let roi_group_data_length = self.read_long()?;
+        let mut nonvar_frame_data = vec![0u8; usize::try_from(nonvar_frame_data_length)?];
+        self.reader.read_exact(&mut nonvar_frame_data)?;
+        let mut roi_group_data = vec![0u8; usize::try_from(roi_group_data_length)?];
+        self.reader.read_exact(&mut roi_group_data)?;
+        // Parse the read bytes into UTF8 and into the metadat object
+        let nonvar_frame_data = String::from_utf8(nonvar_frame_data)?;
+        let roi_group_data = String::from_utf8(roi_group_data)?;
```
HeroicKatora, line 487: If you want these to error with `InvalidScanImageMetadataEncoding` then you can use:

```suggestion
        let nonvar_frame_data = String::from_utf8(nonvar_frame_data)
            .map_err(|_| InvalidScanImageMetadataEncoding)?;
        let roi_group_data = String::from_utf8(roi_group_data)
            .map_err(|_| InvalidScanImageMetadataEncoding)?;
```

HeroicKatora: You could add a field to the variant to keep the underlying conversion error—and display it if you want.

HeroicKatora (changes requested):
I'm quite open to this in principle. The major concern is that it should affect other users in the least amount possible but it seems the inventors of the format did at least apply some amount of diligence of inserting their own data into the format without requiring major changes.

> Another possible blocking issue is that the demo file is a bit big (3 MB I think)

Yes, I think so too. A smaller image of ~100KB would be fine, best would be ~10KB for the pure meta data representation. That said, this is not a major concern and even 3MB would still be acceptable as long as it remains a single test case—but then all of the relevant properties need to be contained in it when you add tests. Please remember, when changing the test image, that you need to rewrite the commit that introduced the 3MB image. Only adding a new commit is not enough to remove the data from `git`.

Farkal:
> 
> 
> I'm quite open to this in principle. The major concern is that it should affect other users in the least amount possible but it seems the inventors of the format did at least apply some amount of diligence of inserting their own data into the format without requiring major changes.
> 
> > Another possible blocking issue is that the demo file is a bit big (3 MB I think)
> 
> Yes, I think so too. A smaller image of ~100KB would be fine, best would be ~10KB for the pure meta data representation. That said, this is not a major concern and even 3MB would still be acceptable as long as it remains a single test case—but then all of the relevant properties need to be contained in it when you add tests. Please remember, when changing the test image, that you need to rewrite the commit that introduced the 3MB image. Only adding a new commit is not enough to remove the data from `git`.

Yeah i don't know if this should be in the main lib, because geotiff could also be integrated here but by providing great abstraction it allow anybody to easily extend it and create custom lib usin image-tiff as core package. Or perhaps we should use some feature flags ?

HagaiHargil:
I resolved all comments, but still haven't gotten new test data so this is on hold for a couple of days :)

Regarding your API comments - I indeed want to keep everything as frictionless as possible for users, but I'm not sure that the added flag is necessary. The chances of the first four bytes being identical are very slim, and I can make the next check also 'blocking' - i.e. make sure that the next bytes after that magic number, which holder the version number, are always an unsigned number between 3 and 10, for example. This will make it even less likely to occur "in the wild" and will make things smoother for the user. 

If you still think opting in is the way to go I'll be happy to do that. I'm not sure why should this be outside of the library - it's a standard TIFF format, albeit pretty similar to BigTIFF, which is great for the guy implementing it, but irrelevant to the users who want their metadata parsed.

HeroicKatora:
> Or perhaps we should use some feature flags?

[/maintainer hat off] With regards to GeoTIFF I would be very happy to see it here as well. It a large use-case and one that could help inform us if the interface is good reading partial data etc. without requiring hacks of `unsafe`.

[/maintainer hat on]. It's not quite as simple. The primary motivations for a separate crate would be:

a) Compile time and binary size but adding a few types shouldn't matter too much. There's also lto removing dead functions. We can use a `feature` as well for the more sizable additions.
b) Interface stability, on which this does well by not interpreting the tags themselves (minus the few suggestions in the review). For GeoTIFF I'm not as sure but it should also be mainly some tags. Requiring the caller to explicity call a different method like the recent modern jpeg reading could work as well.
c) Maintenance work. Doesn't seem too much of a risk here but with GeoTIFF it will be more.

But there are also some advantages:

a) Link time is longer for a large number of fractured crates.
b) Coherence of trait impls and using private methods. This avoids duplicating some functionality.
c) It's easier to discover and the interfaces might be similar.

> I indeed want to keep everything as frictionless as possible for users, but I'm not sure that the added flag is necessary.

In terms of feature flags, my gut feeling is that an opt-in bool is enough but necessary. We shouldn't rely on chance here to assume that no image accidentally has the same magic constant between header and first ift. These are exactly the kind of behaviours that are incredibly hard to debug. When reading a Tiff no one expects that the requirements of another format—they probably never heard of outside your field—are enforced.

HagaiHargil:
@HeroicKatora Got it. The thing is that calling `new()` also calls `init()` which reads the header - and I can't insert `with_scanimage()` in between them, so the data isn't really parsed. Would you like the `with_scanimage()` method to do the actual parsing instead? This means that it's not like `with_limits()`.

Farkal:
I work with geottif and i don't think it's a great idea to have it here. For exemple you will have to add xml lib to be able to decode and encode GdalMetadata, if after that you want to support Cloud Optimized Geotiff you will have to support Ghost headers (that came at the begining of the file and also before each slot of data containing the size of the slot (this prevent the need to store the index and the size of each slot, we only need to store the index and read the 4 bytes before the start of this index to get the size of the slot).  
I prefer to put my code in other lib and make PR here when i found that the interfaces doesn't fit my needs. You can check my geotiff lib here [https://github.com/Farkal/rust_geotiff](https://github.com/Farkal/rust_geotiff) (i use my custom branch of this lib because i added some errors to optimise cloud access).

HagaiHargil:
I uploaded the lightweight SI test file. Tests still fail due to that API issue I discussed earlier.

HeroicKatora:
@HagaiHargil  An alternative would be to make it another constructor like `new`: `with_scanimage(reader: R) -> Reader<R>`. That must be early enough.


> I prefer to put my code in other lib and make PR here when i found that the interfaces doesn't fit my needs.

That's the question, isn't it? Is the current interface enough? In particular, in that it seems to share this with GeoTIFF's Ghost Header, it is a popular option to extend the format by putting data between the header and the first IFD or into the skipped over space before an IFD in general. And even more, both ScanImage and Ghost Header employ practically the same mechanism for this except that ScanImage marks itself with a magic number to make it less prone to misinterpretation. (I can't find any such mechanism for GeoTiff. I might just have missed it.) It definitely is currently tedious to read these by hand. Maybe we could boild this down to a mechanism to read those parts of the file?

If there is a good interface for reading this data then I would also prefer it over the approach here. But for the moment there is not. There is explicitly no policy in the PR, data is left uninterpreted for the most part. If the code is unlikely to change why not have it live here?

@Farkal And a final clarification although slightly off-topic, I meant only to support the tags of GeoTIFF by name and value type. Not that all their interpretation might be moved since there are many choices to make in that implementation.

tests/decode_images.rs
```diff
@@ -183,6 +187,43 @@ fn issue_69() {
...
+/// generate the correct hash value to be compared.
+#[test]
+fn test_si_header_metadata_parser() {
+    let filename = "ScanImageBigTIFFv57.tif";
+    let path = PathBuf::from(TEST_IMAGE_DIR).join(BIGTIFF_DIR).join(filename);
+    let file = File::open(path).unwrap();
+    let decoder = Decoder::new(file).unwrap().with_scanimage();
+    let meta = decoder.scanimage_metadata().unwrap();
+    let mut hasher_roi = DefaultHasher::new();
+    let mut hasher_frame = DefaultHasher::new();
```
HeroicKatora, line 200: The default hasher is not deterministic between runs, by design. Use a crc such as `crc32fast` instead.

tests/decode_images.rs
```diff
@@ -183,6 +187,43 @@ fn issue_69() {
...
+fn generate_new_hash_for_scanimage_metadata_parser() {
+    let filename = "ScanImageBigTIFFv57.tif";
+    let path = PathBuf::from(TEST_IMAGE_DIR).join(BIGTIFF_DIR).join(filename);
+    let file = File::open(path).unwrap();
+    let decoder = Decoder::new(file).unwrap();
+    let meta = decoder.scanimage_metadata().unwrap();
+    let mut hasher_roi = DefaultHasher::new();
+    let mut hasher_frame = DefaultHasher::new();
+    let _ = meta.roi_group_data.hash(&mut hasher_roi);
+    let _ = meta.nonvar_frame_data.hash(&mut hasher_frame);
```
HeroicKatora, line 222: Would you move the calculation into a common method to ensure this skipped test does not silently diverge?

HeroicKatora (changes requested):

HagaiHargil:
I added the `new_with_scanimage` method, but then I realized that the TIF format is `int16`, which `image-tiff` doesn't support. I guess I should implement int16 support first?... :)

HagaiHargil:
So I took a shot at implementing the `int16` functionality here. It can be divided to a different PR if you wish (it's a single squashed commit), but for the moment it's here since it's required for this work.

I wasn't really able to solve all issues - I managed to not break any existing test, but when reading my test file the program crashes due to some OS error. I believe that the offset value that it reads in `initialize_strip_decoder()` is too high, which makes it walk outside of the image file altogether, but I'm not so sure.

Is there any chance you're taking a look at it?

HagaiHargil:
@HeroicKatora Do you think you'll be able to take a look?

HagaiHargil:
I added two `int16` TIFs - one gray and one RGB - and they pass the tests following my changes. I believe that this addressed #113.

Unfortunately the ScanImage-related stuff still doesn't work, I'm guessing I'm doing something wrong on my end - misunderstanding the spec, reading the wrong amount of bytes, etc. What do you want to do @HeroicKatora - will you be able to take a look at these SI issues or do you wish me to separate the `int16` part from the SI part and push that as a separate PR?

HeroicKatora:
I'd definitely prefer `int16` as a separate PR.

PR #103, opened 2020-10-31
WIP: LZW encoding TIFF files

aspenluxxxy:


fintelia:
It looks like this accidentally converted spaces to tabs across a whole bunch of files

aspenluxxxy:
fixed

aspenluxxxy:
why is this building locally but not on CI

fintelia:
The failing CI build was with the 1.34 version of the Rust compiler. It looks like the other ones gets canceled if that fails.

HeroicKatora:
The edition was upgraded to 2018, I believe this should work on 1.34 now.

Issue #106, opened 2020-11-16
"Inconsistent sizes" when parsing LZW compressed images


ruffson:
With _some_ LZW compressed TIFF images, I am getting a `Format error: Inconsistent sizes encountered.` in the `expand_strip(...)` method of `decoder/mod.rs`.

I uploaded 3 example files that show this problem [here](https://cloud.xeres.org/s/QX9z3F5Me8E5RGE) (~630MB).

Last time when I tried to debug this the only thing I noticed is that in the `LZW::new(...)` the `uncompressed.len()` is larger than the buffer size. And iirc the `len()` was the image' width or length.

Anyways, for the three images I printed the `bytes`, `buffer.byte_length()` and the `max_uncompressed_length` right before the error is returned in `expand_strip(...)`:

#### `crater1_lzw.tif`:
bytes: `23714`, buffer_len: `1,` max_uncompressed_length: `23710`
#### `crater2_lzw.tif`
bytes: `23714`, buffer_len: `1`, max_uncompressed_length: `23710`
#### `geotiff2.tif`
bytes: `7710`, buffer_len: `1`, max_uncompressed_length: `2570`

About the images: The first two are from the [CTX data](http://murray-lab.caltech.edu/CTX/index.html) from the Mars Reconnaissance Orbiter. They _should_ be public domain but I am not 100% sure. 
The `geotif2.tif` is taken from the sample files of the [geotiff library](https://github.com/OSGeo/libgeotiff) (and as far as I can tell the official geotiff organization) [here](http://download.osgeo.org/geotiff/samples/gdal_eg/).

One last thing about these samples, if you think that these provide perfect examples of geotiff, check their [readme](http://download.osgeo.org/geotiff/samples/README), specifically: "Just because a file is in this tree does not imply that it follows the standard correctly [...]"

mntns:
I'm facing the same error in `expand_strip()` when trying to load large LZW-compressed images. As already mentioned on gitter, it appears to be a regression from 0.5.0 to 0.6.0. My debugging so far only resulted in locating it in `expand_strip()`, as @ruffson already wrote, but [here is another example image](https://images.wallpen.com/image_tiff_lzw_example.tif) that fails to load with `InconsistentSizesEncountered` (also ~630MB, but a single image).

PR #108, opened 2020-12-03
Fix lzw decoding stopping short of full decode

HeroicKatora:
When the required amount of data was just a tiny bit larger than full a
page size (one buffer increment), the decoder would stop since all
compressed data had been read. However, the lzw decoder contains a small
internal bit buffer for code decoding. That is, some remaining bytes had
been retained and were still left to be decoded.

Analysis was complicated by a (probably) incorrect sign in a sanity
check: We had a check that the decompressed bytes were at most as large
as required. Instead, we need at least as much data as required while an
internal invariant should stop at the request size, but it's not a
critical format error. This has been replaced with more concise errors.

Closes: #106

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -354,6 +354,11 @@ impl<R: Read + Seek> Decoder<R> {
             PhotometricInterpretation::RGB => match self.bits_per_sample[..] {
                 [r, g, b] if [r, r] == [g, b] => Ok(ColorType::RGB(r)),
                 [r, g, b, a] if [r, r, r] == [g, b, a] => Ok(ColorType::RGBA(r)),
+                // FIXME: We should _ignore_ other components. In particular:
+                // > Beware of extra components. Some TIFF files may have more components per pixel
+                // than you think. A Baseline TIFF reader must skip over them gracefully,using the
+                // values of the SamplesPerPixel and BitsPerSample fields.
+                // > -- TIFF 6.0 Specification, Section 7, Additional Baseline requirements.
                 _ => Err(TiffError::UnsupportedError(
                     TiffUnsupportedError::InterpretationWithBits(
                         self.photometric_interpretation,
@@ -766,7 +771,7 @@ impl<R: Read + Seek> Decoder<R> {
         buffer: DecodingBuffer<'a>,
         offset: u64,
         length: u64,
-        max_uncompressed_length: usize,
+        strip_sample_count: usize,
     ) -> TiffResult<usize> {
         let color_type = self.colortype()?;
         self.goto_offset_u64(offset)?;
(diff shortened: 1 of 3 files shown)


src/decoder/mod.rs
```diff
@@ -804,9 +804,12 @@ impl<R: Read + Seek> Decoder<R> {
             }
         };
 
-        if bytes / buffer.byte_len() > max_uncompressed_length {
+        if bytes / buffer.byte_len() < max_uncompressed_length {
```
HeroicKatora, line 807: I'm not 100% sure about this yet. It seems illogical to check if we had read _more_ data but currently we avoid this in `lzw` at least.

HeroicKatora (commented):
Still looking for a small regression test. @ruffson @mntns
(The 600MB image passes though)

src/decoder/mod.rs
HeroicKatora: This reveals an interesting test failure. `u12` data is packed in bytes but the decoder uses the `buffer.byte_len()` and multiplies it with the sample count to determine the expected buffer size. That's incorrect. Instead, the strip's data representation should be independent of the BitsPerSample.

fintelia: I think that the original code for this line was correct. Specifically for reasons I don't really understand, `max_uncompressed_length` is being used as the size for `buffer`. This check is to make sure that the subsequent indexing operations on buffer do not panic.

The case of too little data being read is caught by the calling function https://github.com/image-rs/image-tiff/blob/c9703dcb8a008567c5d9c6263195f56850e253a7/src/decoder/mod.rs#L1038

HeroicKatora: Hm, [this Stackexchange answer](https://stackoverflow.com/questions/7068183/writing-10-12-bit-tiff-files-with-libtiff-c) suggest that while libTIFF packs the bits, there is no actually standard prescribed behavior. We might have to adjust for that?

fintelia: Isn't any `u12` or similar data going to trigger UnsupportedColorType error below?

HeroicKatora: The problem is, this triggers _first_ and if the other change is reverted than we must check the units_read.

src/error.rs
```diff
@@ -43,6 +43,14 @@ pub enum TiffFormatError {
     TiffSignatureInvalid,
     ImageFileDirectoryNotFound,
     InconsistentSizesEncountered,
+    UnexpectedCompressedData {
+        actual: usize,
+        required: usize,
```
fintelia, line 48: Perhaps actual_bytes and required_bytes?

src/error.rs
```diff
@@ -43,6 +43,14 @@ pub enum TiffFormatError {
...
+    },
+    InconsistentStripElements {
+        actual: usize,
+        required: usize,
```
fintelia, line 52: Similar to above, perhaps actual_elements and required_elements?

fintelia (approved):

PR #110, opened 2020-12-04
Refactor Entry::val function

### LINKED ISSUE #109: why does tag XResolution return a List?
After updating from 0.5.0 to 0.6.0 I got a pattern matching error. It appear the tag XResolution in the decoder gives a List, whil before it did not:

decoder.find_tag(Tag::XResolution)?
Returns in 0.6.0:
Some(List([Rational(250000000, 250000000)]))

In 0.5.0 it returns:
Some(Rational(250000000, 250000000))

Is this intentional?

fintelia:
This change cleans up the logic for reading tag values. In the process, it should fix #109.

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -7,7 +7,7 @@ use std::mem;
 
 use super::stream::{ByteOrder, EndianReader, SmartReader};
 use crate::tags::{Tag, Type};
-use crate::{TiffError, TiffFormatError, TiffResult, TiffUnsupportedError};
+use crate::{TiffError, TiffFormatError, TiffResult};
 
 use self::Value::{
     Ascii, Byte, Double, Float, List, Rational, RationalBig, SRational, SRationalBig, Short,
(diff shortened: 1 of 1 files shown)


src/decoder/ifd.rs
```diff
@@ -332,171 +332,191 @@ impl Entry {
...
-        } else {
-            match (self.type_, self.count) {
-                // TODO check if this could give wrong results
-                // at a different endianess of file/computer.
-                (Type::BYTE, 1) => Ok(Unsigned(u32::from(self.offset[0]))),
-                (Type::BYTE, 2) => offset_to_bytes(2, self),
-                (Type::BYTE, 3) => offset_to_bytes(3, self),
-                (Type::BYTE, 4) => offset_to_bytes(4, self),
-                (Type::BYTE, n) => self.decode_offset(n, bo, limits, decoder, |decoder| {
-                    Ok(UnsignedBig(u64::from(decoder.read_byte()?)))
```
fintelia, line 417: It isn't clear to me why this returns UnsignedBig instead of Byte, but we should probably make a decision about whether to retain that going forward.

src/decoder/ifd.rs
```diff
@@ -332,171 +332,191 @@ impl Entry {
         limits: &super::Limits,
         decoder: &mut super::Decoder<R>,
     ) -> TiffResult<Value> {
+        // Case 1: there are no values so we can return immediately.
+        if self.count == 0 {
+            return Ok(List(Vec::new()));
```
fintelia, line 337: Not sure if an empty list is the right thing to return, but that's what the code was doing before.

src/decoder/ifd.rs
HeroicKatora: Possibly because `Byte` is newer. I'm not 100% sure about that though.

src/decoder/ifd.rs
```diff
@@ -332,171 +332,206 @@ impl Entry {
...
+        // Case 3: There is more than one value, but it fits in the offset field.
+        if value_bytes <= 4 || decoder.bigtiff && value_bytes <= 8 {
+            match self.type_ {
+                Type::BYTE => return Ok(offset_to_bytes(self.count as usize, self)?),
+                Type::SBYTE => return Ok(offset_to_sbytes(self.count as usize, self)?),
+                Type::ASCII => {
+                    let mut buf = vec![0; self.count as usize];
+                    self.r(bo).read_exact(&mut buf)?;
+                    let v = String::from_utf8(buf)?;
+                    let v = v.trim_matches(char::from(0));
```
HeroicKatora, line 424: Interesting. This also comes from the previous implementation but reviewing it now it seems to be at odds with the validating of the one-byte case. It does _not_ validate the the last byte is indeed NUL and also removes NUL from both sides of the string. Not blocking this PR but we should open an issue on it?

HeroicKatora (approved):

src/decoder/ifd.rs
HeroicKatora: Let's just keep the case here for compatibility. I suppose we should rework the `Value`-system anyways. The allocation can not be re-used and not be done upfront because every different type requires a different `Vec<_>` allocation, instead of a generically aligned byte buffer.

Issue #107, opened 2020-12-02
Support strip size > 8kB


tnl-rwa:
In ImageEncoder::new() the strip size of the image data in the tiff is limited to 8kB, with the remark it is per tiff spec. But the tiff spec of 1992(!!) states a recommendation for the 8kB as mitigation for out of memory errors. At time has passed 8kB is ridiculous small. 
I bumped into this as I faced performance issues in the Python Pillow library. We have an application retrieving rectangular parts from large images. I noticed that the performance of images generated via the image-tiff crate was ten time slower than images created by Pillow itself. It appeared that Pillow stored the image as a single strip, whil image-tiff stored one line per strip.

I would like to see an option to store the image in a single strip, configure the max strip size or even get rid of the 8kB strip limit. 

https://github.com/image-rs/image-tiff/blob/master/src/encoder/mod.rs#L625

When agreed on a solution, I'll implement it myself.

aschampion:
Just to note what Python's `tifffile` (which is a great lib) does [here](https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#L1203-L1206):

```
        rowsperstrip : int
            The number of rows per strip. By default, strips are ~64 KB if
            compression is enabled, else rowsperstrip is set to the image
            length.
```

I agree it should be configurable with more sensible defaults that handle the typical uses of one plane-per-strip and strip'd but modernly sized (64KB seems nice), e.g., by a config enum of ContiguousPlanes/RowsPerStrip(u32)/StripSize(u32)

tnl-rwa:
You are doing a great job on the TIFF library in Rust. Thank you very much for pulling my request in the archive.
Will there be a version 0.6.1 soon, containing the latest fixes? I would like to integrate it into my project.

HeroicKatora:
On Friday or the weekend most likely. If it's pressing you can use a [`git` source for the dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).

PR #111, opened 2020-12-05
Large strip size

tnl-rwa:
Closes: #107

### DIFF
--- a/src/encoder/mod.rs
+++ b/src/encoder/mod.rs
@@ -760,7 +760,7 @@ impl<'a, W: 'a + Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
 
     /// Set image number of lines per strip
     ///
-    /// This function must be called before calling write_data() or write_strip().
+    /// `write_strip` and will return an error otherwise.
     pub fn rows_per_strip(&mut self, value: u32) -> TiffResult<()> {
         if self.strip_idx != 0 {
             return Err(io::Error::new(
--- a/src/encoder/mod.rs
+++ b/src/encoder/mod.rs
@@ -589,6 +589,9 @@ impl<'a, W: Write + Seek> Drop for DirectoryEncoder<'a, W> {
 /// // You can encode tags here
 /// image.encoder().write_tag(Tag::Artist, "Image-tiff").unwrap();
 ///
+/// // Smaller strip size can be configured before writing data
+/// image.rows_per_strip(2).unwrap();
+///
 /// let mut idx = 0;
 /// while image.next_strip_sample_count() > 0 {
 ///     let sample_count = image.next_strip_sample_count() as usize;
@@ -620,10 +623,9 @@ impl<'a, W: 'a + Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
         height: u32,
     ) -> TiffResult<ImageEncoder<'a, W, T>> {
         let row_samples = u64::from(width) * u64::try_from(<T>::BITS_PER_SAMPLE.len())?;
-        let row_bytes = row_samples * u64::from(<T::Inner>::BYTE_LEN);
 
-        // As per tiff spec each strip should be about 8k long
-        // let rows_per_strip = (8000 + row_bytes - 1) / row_bytes;
+        // By default write the image as single strip. This can be changed by calling
+        // rows_per_strip() before the first data is written.
         let rows_per_strip = height as u64;
 
         let strip_count = (u64::from(height) + rows_per_strip - 1) / rows_per_strip;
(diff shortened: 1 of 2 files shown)


tests/encode_images.rs
```diff
@@ -360,3 +360,19 @@ fn test_multipage_image() {
...
+    let mut img_encoder = TiffEncoder::new(&mut file).unwrap();
+
+    let mut image = img_encoder.new_image::<colortype::Gray8>(100, 100).unwrap();
+    assert_eq!(image.next_strip_sample_count(), 100 * 100);
+    image.rows_per_strip(2).unwrap();
+    assert_eq!(image.next_strip_sample_count(), 2 * 100);
+
+    let img2: Vec<u8> = vec![0; 2 * 100];
+    image.write_strip(&img2[..]).unwrap();
+    assert!(image.rows_per_strip(5).is_err());
```
HeroicKatora, line 377: Please add a roundtrip test that tries to decode the written image.

src/encoder/mod.rs
```diff
@@ -755,6 +758,27 @@ impl<'a, W: 'a + Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
...
+    pub fn rows_per_strip(&mut self, value: u32) -> TiffResult<()> {
+        if self.strip_idx != 0 {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "Cannot change strip size after data was written",
+            )
+            .into());
+        }
+        // Write tag as 32 bits
+        self.encoder.write_tag(Tag::RowsPerStrip, value)?;
```
HeroicKatora, line 773: This tag is already written in `new`. Is that a problem?

HeroicKatora: Calling `new_image` will directly construct the `ImageDecoder::new` and write the tags. Maybe it would be helpful to have an intermediate build structure, or another method with different argument type than the sizes.

src/encoder/mod.rs
```diff
@@ -620,10 +623,10 @@ impl<'a, W: 'a + Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
         height: u32,
     ) -> TiffResult<ImageEncoder<'a, W, T>> {
         let row_samples = u64::from(width) * u64::try_from(<T>::BITS_PER_SAMPLE.len())?;
-        let row_bytes = row_samples * u64::from(<T::Inner>::BYTE_LEN);
 
-        // As per tiff spec each strip should be about 8k long
-        let rows_per_strip = (8000 + row_bytes - 1) / row_bytes;
```
fintelia, line 626: I think it would be better to keep the default in terms of number of bytes, but just significantly increase the target. Maybe 256KB or 1MB or so?

src/encoder/mod.rs
tnl-rwa: As far as I can see and test (see updated test), this is not a problem.
write_tag() does not write to file directly. It writes the tags into a map in memory. When a tag already exists  in the map, the value is overwritten with the new value. The map is written to file later when finish is called.

src/encoder/mod.rs
tnl-rwa: I checked multiple libraries and see the default for RowsPerStrip is the image height. What would be a good reason for using a custom arbitrary default?

tests/encode_images.rs
tnl-rwa: I have extended the test with a round-trip

src/encoder/mod.rs
```diff
@@ -755,6 +758,27 @@ impl<'a, W: 'a + Write + Seek, T: ColorType> ImageEncoder<'a, W, T> {
         self.encoder.write_tag(Tag::YResolution, value).unwrap();
     }
 
+    /// Set image number of lines per strip
+    ///
+    /// This function must be called before calling write_data() or write_strip().
```
HeroicKatora, line 763:
```suggestion
    /// This function needs to be called before any calls to `write_data` or `write_strip` and will return an error otherwise.
```

tests/encode_images.rs
```diff
@@ -360,3 +360,42 @@ fn test_multipage_image() {
...
+        let mut decoder = Decoder::new(&mut file).unwrap();
+        assert_eq!(decoder.get_tag_u64(Tag::RowsPerStrip).unwrap(), 2);
+        assert_eq!(decoder.strip_count().unwrap(), 50);
+
+        for i in 0..50 {
+            let img2 = [i; 2 * 100];
+            if let DecodingResult::U8(data) = decoder.read_strip().unwrap() {
+                assert_eq!(&img2[..], &data[..]);
+            } else {
+                assert!(false);
```
HeroicKatora, line 397:
```suggestion
            match decoder.read_strip().unwrap() {
                DecodingResult::U8(data) => assert_eq!(&img2[..], &data[..]),
                other => panic!("Incorrect strip type {:?}", other),
```

HeroicKatora (commented):
Just two stylistic suggestions.

src/encoder/mod.rs
HeroicKatora: Indeed, for example ImageMagick uses that strategy. On the 'why not' though, it appears to have been the cause for [at least one denial-of-service weakness](https://github.com/ImageMagick/tiff/blob/19f25769da6bde2434e89305e7abbf2abf729996/html/v4.0.9.html#L221-L224). I'd rather have some limit here as well, and be it just to keep the expectation of handling multiple strips 'oiled' so to speak. Maybe the default limit we use for our decode buffer, that is 256 MB? That should hardly limit decoding throughput for quite some time while still being reasonably small in terms of total memory—even though *we* don't allocate the caller will and it's nice to nudge them in the right direction of considering OOM.

fintelia: For compressed images, you can only seek to the start of a strip. And decompressing large images can take a non-trivial amount of time so seeking to read only the part you need can be quite useful useful.

tnl-rwa: Ok, then I'll set it to default 1MB strip size instead of the original 8KB. Agree?

HeroicKatora: Sounds good to me.

HeroicKatora (approved):

tnl-rwa:
@HeroicKatora Can this pull request already be merged, or is something else to be done?

PR #114, opened 2020-12-08
Int8/16 support

HagaiHargil:
This PR adds support for TIF files with a datatype of `int16`. Two `int16` files - gray and RGB - were generated using Python's `tifffile` and the decoding results here match theirs.


src/decoder/mod.rs
```diff
@@ -899,6 +932,16 @@ impl<R: Read + Seek> Decoder<R> {
...
                 bytes / 2
             }
+            // TODO: not sure about the following implementation
+            (ColorType::Gray(16), DecodingBuffer::I16(ref mut buffer)) => {
+                reader.read_i16_into(&mut buffer[..bytes / 2])?;
+                if self.photometric_interpretation == PhotometricInterpretation::WhiteIsZero {
+                    for datum in buffer[..bytes / 2].iter_mut() {
+                        *datum = 0xff - *datum
+                    }
+                }
```
HeroicKatora, line 942: Let's just report this case as unsupported for now? Unless we have any test file that actually executes the case I wouldn't feel comfortable guessing the semantics.

src/decoder/mod.rs
```diff
@@ -1112,6 +1151,12 @@ impl<R: Read + Seek> Decoder<R> {
                     TiffUnsupportedError::UnsupportedBitsPerChannel(n),
                 )),
             },
+            SampleFormat::Int => match max_sample_bits {
+                n if n <=16 => DecodingResult::new_i16(buffer_size, &self.limits),
```
HeroicKatora, line 1155: The `n <= 8` case should be treated separately. Some code depends on the type of samples to evaluate the expected number of bytes. You could test that case by producing an `int8_rgb` image.

HeroicKatora (commented):
Thanks for adding test files right away. When you say, 'the result matches theirs', how did you determine this?

HagaiHargil:
> Thanks for adding test files right away. When you say, 'the result matches theirs', how did you determine this?

I should've been clearer. I generated an array in numpy, serialized it with `tifffile` and verified that after decoding with `image-tiff` the values are the same. I look at the starting values, ending ones and (obviously) the total sum which is shown in the test.

src/decoder/mod.rs
fintelia: Agreed. 0xff seems like the wrong value to use, but also "white is zero" isn't a very meaningful thing to say about integers that range from -32768 to 32767.

src/decoder/mod.rs
```diff
@@ -1047,21 +1088,19 @@ impl<R: Read + Seek> Decoder<R> {
         let strip_height = cmp::min(rows_per_strip, sized_height - index * rows_per_strip);
 
         let buffer_size = sized_width * strip_height * self.bits_per_sample.len();
-
+
```
fintelia, line 1091: nit: remove extra whitespace here

src/decoder/mod.rs
fintelia: Actually looking at the libtiff source code, I think that the variants actually mean "min is white" and "max is white". In which case, the correct conversion would be:
```rust
*datum = -1 - *datum;
```

HagaiHargil:
Hi, sorry for disappearing, been a tough few days at work. I'm happy you helped with styling, sorry for forgetting about `rustfmt`. You had one major comment which I really didn't get to the bottom of, so instead of thinking too hard I simply added `int8` support as well.

HeroicKatora (approved):

fintelia (approved):
This looks good to me. It is a breaking change since it adds variants to DecodingResult, so we may want to think about doing the same for int32 and int64 before publishing a new version

HeroicKatora:
We'll want to look at a more future-proof way anyways. The last two breaking releases (and then the next) or so have been motivated by similar issues.

PR #118, opened 2021-02-20
Open tif files containing i32 arrays

satarsa:
I use image-tiff to open frames which are produced by various x-ray detectors. It worked fine until I met a tiff file from Dectris Pilatus detector. This detector uses negative values to mask dead pixels, and, consequently it stores i32 arrays in tiff file. A sample file can be downloaded from here:
https://cloud.esrf.fr/s/i4YLE9Pyb6dHGES
I wrote a trivial patch which allows to open those tiffs. Let me know how I can make it better.


fintelia:
This looks good to me, though might be worth adding support for i64 at the same time

HeroicKatora (approved):
Same, though I can also accept this as is since apparently writing `i64` is also not supported but it is for `i32` (see `writer.rs`, `impl TiffValue for [i32]`).

PR #121, opened 2021-03-20
Validation for ASCII tag type

### LINKED ISSUE #102: ASCII tag type is not validated enough
The specification says the following:

> 2 = ASCII 8-bit byte that contains a 7-bit ASCII code; the last bytemust be NUL (binary zero)

We instead validate that the string is valid UTF-8, a superset of the allowed values

https://github.com/image-rs/image-tiff/blob/679f4f075ba653a29de394e4c6413f94d2933b9c/src/decoder/ifd.rs#L478-L480

This may not seem problematic at first glance but callers might assume that the string can be sliced at any byte. This would panic at runtime if the index turned out not to be a utf-8 character boundary. If values were validated to be truly ascii then all byte boundaries are valid character boundaries.

emmanuelantony2000:
Closes #102

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -4,6 +4,7 @@ use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use std::io::{self, Read, Seek};
 use std::mem;
+use std::str;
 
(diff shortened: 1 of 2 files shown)


HeroicKatora (approved):

HeroicKatora:
Thank you!

PR #122, opened 2021-03-23
Add support for encoding BigTiff

phil-opp:
This pull request implements basic encoding support for BigTiff images by making the encoder types generic over a new `TiffKind` trait.

TODO:

- [ ] Do we need to adjust `pad_word_boundary`?
- [ ] Right now users need to explicitly specify whether they want to create a Tiff or BigTiff file. It might be also possible to (roughly) derive this from the image dimensions and color type. However, I'm not sure if that's a good idea since BigTiff is not supported everywhere.
- [ ] This change is technically breaking because it introduces a new generic parameter to the encoder types. For the main `TiffEncoder` type I made this parameter optional by setting a default, so there should not be any breakage through this. However, if people use the `DirectoryEncoder` or `ImageEncoder` types explicitly, it might lead to breakage. Is this ok or should I try to avoid breakage there as well? 

cc #3 @Farkal


Farkal:
I think we need this internal trait to be able to implement different encoding easily (same for reading). For example the cloud optimized geotiff encoding could be easy to implement with this trait.
We could add method that directly set the generic parameter to avoid setting it again and again.

phil-opp:
I'm happy to try a different approach if you prefer. I'm not sure if a boolean (or enum) parameter would make things easier though since it also needs to be passed around (e.g. to `DirectoryEncoder` and `ImageEncoder`), but I can try it if you like.

I agree that exposing the internal trait in the interface is not ideal. On the other hand, you would not need to specify them as a user in most situations. One way to hide them completely would be to create separate `TiffEncoder` and `BitTiffEncoder` wrapper types that internally dispatch to the generic version.

HeroicKatora:
If the internal trait were object-safe then it would be entirely possible to add multiple implementations behind a `Box<dyn TiffKind>`. I realize this requires a bit of rework since associated types can't be part of the interface (it would make it non-object-safe) but if the implementation is actually stateless then the boxed zero-sized-types wouldn't even require a real allocation.

phil-opp:
Looking through the code again, I still think that the `TiffKind` trait is the better solution, compared to a boolean/enum field. The problem is that such a field would be required in the `DirectoryEncoder::new` and `ImageEncoder::new` constructors as well.

I pushed a new commit to improve the documentation a bit and I think the interface is still easily usable, even with the `TiffKind` trait:

> image
> 
> image
 
I see the `TiffKind` trait similar to the `TiffValue` trait: They are not intended to be implemented in other crates, but they document what kind of values are possible, so they are useful to have in the public interface:

> image

HeroicKatora:
Okay, I'm not going to stand in the way if this makes the encoder work :+1:

phil-opp:
I integrated the latest changes from master to resolve the merge conflicts. We tested this PR with some example images and it seems to work fine, so it's ready for review from my side. Note that there is still an open `TODO` about `pad_word_boundary` that I'm not sure about.

Farkal:
It seems good to me 😉 
Don't worry for the `pad_word_boundary` we can add method after if we need it.
I let @HeroicKatora merge if he is also ok 😉

HeroicKatora (approved):

HeroicKatora:
I'm not quite sure if there is any other breaking change in progress or prepared, ping me on a release PR bumping versions and changelog if you want to release of this state.

Issue #124, opened 2021-04-04
Erroneous JPEG decoding


TomasKralCZ:
Some JPEG-in-TIFF images aren't correctly decoded.

For example: (tested with Emulsion 8.0.0)
![jpegdecoding](https://user-images.githubusercontent.com/23663355/113514260-e1f78100-956d-11eb-9bbf-fb1bb8811f32.png)

Output from tiffdump:
```
stripped-jpeg.tif:
Magic: 0x4949 <little-endian> Version: 0x2a <ClassicTIFF>
Directory 0: offset 94702 (0x171ee) next 0 (0)
ImageWidth (256) SHORT (3) 1<365>
ImageLength (257) SHORT (3) 1<547>
BitsPerSample (258) SHORT (3) 3<8 8 8>
Compression (259) SHORT (3) 1<7>
Photometric (262) SHORT (3) 1<2>
FillOrder (266) SHORT (3) 1<1>
StripOffsets (273) LONG (4) 1<8>
Orientation (274) SHORT (3) 1<1>
SamplesPerPixel (277) SHORT (3) 1<3>
RowsPerStrip (278) SHORT (3) 1<547>
StripByteCounts (279) LONG (4) 1<94694>
XResolution (282) RATIONAL (5) 1<72>
YResolution (283) RATIONAL (5) 1<72>
PlanarConfig (284) SHORT (3) 1<1>
ResolutionUnit (296) SHORT (3) 1<2>
PageNumber (297) SHORT (3) 2<0 1>
Whitepoint (318) RATIONAL (5) 2<0.3127 0.329>
PrimaryChromaticities (319) RATIONAL (5) 6<0.64 0.33 0.3 0.6 0.15 0.06>
JPEGTables (347) UNDEFINED (7) 289<0xff 0xd8 0xff 0xdb 00 0x43 00 0x8 0x6 0x6 0x7 0x6 0x5 0x8 0x7 0x7 0x7 0x9 0x9 0x8 0xa 0xc 0x14 0xd ...>
ICC Profile (34675) UNDEFINED (7) 3144<00 00 0xc 0x48 0x4c 0x69 0x6e 0x6f 0x2 0x10 00 00 0x6d 0x6e 0x74 0x72 0x52 0x47 0x42 0x20 0x58 0x59 0x5a 0x20 ...>
```

Same error appeared while I was trying to decode tiled-based images.

[...]

PR #125, opened 2021-04-04
Added basic tile decoding

TomasKralCZ:
Hi there ! This is my first PR to an open-source Rust repo after quite some time so bear with me please :-)

Tile-based decoding isn't much different from strip-based, it's just a matter of trying to fit it nicely into the existing codebase, which I tried really hard to do, but I didn't really come up with elegant code... I basically had to duplicate the whole `expand_strip` method.

I didn't add incremental decoding, because I wasn't sure about the API, but that can be easily done later.

I left a few TODOs in the code. Firstly, I need a good way of skipping some bytes from the Reader to skip padded tiles, but there doesn't seem to be a good way of doing it, so I just allocated one Vec, that I reuse for the whole image.

The other TODO is about errors. I'm not sure how pedantic the decoder should be. For example, the spec says "do **not** use both strip-oriented fields and tile-oriented fields in the same image", but I noticed that some libtiff test images contain both.

I also tried to add tile support for the JPEG decoder, but the JPEG implementation doesn't work for my test images https://github.com/image-rs/image-tiff/issues/124#issue-849904971, so I'll wait until that gets fixed.


src/decoder/mod.rs
```diff
@@ -220,6 +220,30 @@ struct StripDecodeState {
...
+    tiles_down: usize,
+    tiles_across: usize,
+    padding_right: usize,    // Length of padding for rightmost tiles in pixels
+    padding_down: usize,     // length of padding for bottommost tile in pixels
+    padding_buffer: Vec<u8>, // A simple buffer right paddding is read into
+    tile_samples: usize,
+    line_samples: usize,       // Sample count of one row of the image
+    row_samples: usize,        // Sample count of one row of one tile
+    tile_strip_samples: usize, // Sample count of one row of tiles
+}
```
HeroicKatora, line 239: Comments should be doc-comments above the fields, for example:

```
     /// Length of padding for rightmost tiles in pixels
    padding_right: usize,
```

src/decoder/mod.rs
```diff
@@ -1107,6 +1395,41 @@ impl<R: Read + Seek> Decoder<R> {
...
+                tile_offsets: self.get_tag_u64_vec(Tag::TileOffsets)?,
+                tile_bytes: self.get_tag_u64_vec(Tag::TileByteCounts)?,
+                tile_width,
+                tile_length,
+                tiles_across,
+                tiles_down,
+                tile_samples: tile_sample_count,
+                padding_right,
+                padding_down: (tiles_down * tile_length) - usize::try_from(self.height)?,
+                padding_buffer: vec![0; padding_right * samples_per_pixel * buffer_byte_len],
```
HeroicKatora, line 1423: This is allocating a potentially very large buffer where the size is controlled by the input. Please compare the desired length against the `limits`.

HeroicKatora (commented):
I think most of this looks very solid. Where are the images from, so as not to run into any copyright issues on work committed to git? (Their size is okay).

> Firstly, I need a good way of skipping some bytes from the Reader to skip padded tiles, but there doesn't seem to be a good way of doing it, so I just allocated one Vec, that I reuse for the whole image.

This is fine by me. I'd suppose we could add something to the `EndianReader`  here at some point. You could open an issue about this inefficiency.

> I'm not sure how pedantic the decoder should be.

A good baseline is to try and accept anything that other libraries deal with, if there is no ambiguity. There's a larger question of how to report such warnings or violations to the user. In any case, one justification in the code and a `FIXME` comment is good enough, we can always relax this later if it appears to be an issue in practice.

src/decoder/mod.rs
TomasKralCZ: Sure ! I assume the use case for this is running `cargo doc` with `--document-private-items` ?

HeroicKatora: Yep, and because it's just stylistically more common and uniform.

TomasKralCZ:
I don't like that I duplicated some code in the expand_strip/tile methods, so I'll try to refactor that.

The new test photos are mine, so there shouldn't be any problem about that.

How do you think the incremental API should work ?
The strip incremental API lets you read a few strips on their own and then read the rest of the image at once (at least I think that's how it works). The problem with this API for Tiles is that if the user doesn't read all of the tiles in a row, then they would get like an L-shaped result, which probably doesn't make sense. So I'd let the user read one tile at the time, but make it so the next call to read_image() returns the whole image.

HeroicKatora:
> How do you think the incremental API should work ?

What do you think about actually returning an _error_ when calling `read_strip` on a tiled image? The method is about the ability to efficiently access the underlying encoding and I'd argue it is the resposibility of the caller to ensure that this is the correct encoding. For tiles you could add a similar `read_tile` method.

TomasKralCZ:
Sure, I added new error types for that.

Btw, I'm happy to see that @fintelia is working on simplifying the expand_strip function. I'll probably wait until that gets merged and try to mirror that for expand_tile.

HeroicKatora:
@TomasKralCZ Both of these have been merged, so you can reuse the refactored methods :)

TomasKralCZ:
I think the code looks pretty good now.
One thing I'm not sure about is why @fintelia removed color inversion for signed formats. Is that non-standard behaviour ? I remember there was a comment about how libtiff interprets that in a certain way but it got removed.

HeroicKatora (approved):

fintelia (approved):
There's some duplicated code that could probably be factored out, but that can happen in a followup.

PR #126, opened 2021-04-10
Simplify de-compression logic

fintelia:
Behavior should be equivalent, but makes the code somewhat cleaner.


Farkal (approved):

HeroicKatora:
@Farkal Thanks for the reviews

Issue #129, opened 2021-04-18
Color inversion for signed gray images


fintelia:
From @TomasKralCZ in https://github.com/image-rs/image-tiff/pull/125#issuecomment-819480657
>  One thing I'm not sure about is why @fintelia removed color inversion for signed formats. Is that non-standard behaviour ? I remember there was a comment about how libtiff interprets that in a certain way but it got removed.

#128 wasn't intended to change this behavior, but looking more closely it seems that the library wasn't being consistent prior to the change. In particular, [i32](https://github.com/image-rs/image-tiff/blob/9b94e9de8827953dddf255c5a7b32316a2c0ccc3/src/decoder/mod.rs#L1015-L1018) and [i64](https://github.com/image-rs/image-tiff/blob/9b94e9de8827953dddf255c5a7b32316a2c0ccc3/src/decoder/mod.rs#L1002-L1005) grayscale images never did inversion, while [i8](https://github.com/image-rs/image-tiff/blob/9b94e9de8827953dddf255c5a7b32316a2c0ccc3/src/decoder/mod.rs#L1028-L1036) and [i16](https://github.com/image-rs/image-tiff/blob/9b94e9de8827953dddf255c5a7b32316a2c0ccc3/src/decoder/mod.rs#L1039-L1047) would if WhiteIsZero was specified.

We should figure out which is correct behavior and then do it consistently. This comment suggests that inversion might be correct:

https://github.com/image-rs/image-tiff/blob/9b94e9de8827953dddf255c5a7b32316a2c0ccc3/src/decoder/mod.rs#L1037-L1038

TomasKralCZ:
I tried figuring out how libtiff handles this, but the only relevant piece of code I found is this:
<details>
/*
 * Greyscale images with less than 8 bits/sample are handled
 * with a table to avoid lots of shifts and masks.  The table
 * is setup so that put*bwtile (below) can retrieve 8/bitspersample
 * pixel values simply by indexing into the table with one
 * number.
 */
static int
makebwmap(TIFFRGBAImage* img)
{
    TIFFRGBValue* Map = img->Map;
    int bitspersample = img->bitspersample;
    int nsamples = 8 / bitspersample;
    int i;
    uint32_t* p;

    if (nsamples == 0)
        nsamples = 1;

    img->BWmap = (uint32_t**)_TIFFmalloc(
        256 * sizeof(uint32_t*) + (256 * nsamples * sizeof(uint32_t)));
[...]

PR #130, opened 2021-04-21
Prepare 0.7.0 release

phil-opp:
Prepares a new crates.io release with the latest changes as suggested by @HeroicKatora in https://github.com/image-rs/image-tiff/pull/122#issuecomment-817708366. I tried to include all changes since the v0.6.1 release in the changelog. Let me know if I missed anything!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,9 @@ Changes:
   * Switch to streaming decompression via `flate2`. Aside from performance improvements and lower RAM consumption, this fixes a bug where `max_uncompressed_length` was precalculated for a single tile but then used as a hard limit on the whole data, failing to decompress any tiled images.
   * Add support for new `Deflate` tag in addition to `OldDeflate`.
 * _Breaking:_ Remove `InflateError`, which is no longer needed with `flate2` ([#134](https://github.com/image-rs/image-tiff/pull/134))
+* _Breaking:_ Support for `MinIsWhite` is restricted to unsigned and floating
+  point values. This is expected to be be re-added once some contradictory
+  interpretation regarding semantics for signed values is resolved.
 
 Fixes:
 * Validate that ASCII tags are valid ASCII and end with a null byte ([#121](https://github.com/image-rs/image-tiff/pull/121))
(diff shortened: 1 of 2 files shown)


fintelia:
#129 should probably be resolved before this release so we don't cause any any regressions.

HeroicKatora:
Considering:
* This is already taking a while
* We have a somewhat clear plan to address the regression
* There are quite a few extra features in master

I think it might be better to publish `0.7`—clearly stating the caveat in its release notes—and push it as `0.7.1` afterwards.

fintelia (approved):
Let's make a note about changes to MinIsWhite and then I think this is ready to merge.

CHANGES.md
```diff
@@ -1,3 +1,25 @@
+# Version 0.7.0
+
+New features:
+* Support for encoding BigTiff ([#122](https://github.com/image-rs/image-tiff/pull/122))
```
HeroicKatora, line 4: Needs a rebase on `master` and noting the support for new `Deflate` (https://github.com/image-rs/image-tiff/pull/132).

HeroicKatora (changes requested):

CHANGES.md
HeroicKatora: Or not? we can release this as is of course :)

fintelia: Yeah. Probably is also worth pointing out the switch to `flate2`.

RReverser: Actually, since #132 gets here in a 0.7.0 and breaking changes are allowed, perhaps I should remove `InflateError` altogether now?

As part of #132 I left it but marked as deprecated only to avoid breaking API changes.

RReverser: Created PR in https://github.com/image-rs/image-tiff/pull/134.

phil-opp:
I rebased on top of `master` and included the changes of #132 and #134.

> Let's make a note about changes to MinIsWhite

I assume this is about #129? I didn't follow that issue closely, so maybe someone could provide a suggestion for this note?

RReverser:
Sorry to ping, but... is there anything else blocking this release?

HeroicKatora (approved):

fintelia (approved):

HeroicKatora:
@RReverser Thanks for the ping.

PR #132, opened 2021-05-08
Improve deflate support

### LINKED ISSUE #131: Compression method Deflate is unsupported
I'm trying to use `image` crate to read TIFF images produced (compressed) by ImageMagick with `-compress zip`. My understanding is that `image` crate relies on this one to do actual reading, so raising issue here.

This is what Exiftool reports on those images:

```
> exiftool -compression pleiades4.tif
Compression                     : Adobe Deflate
```

README suggests that decoding Deflate TIFFs should be supported:

image

However, when trying to read them using `image::open`, I'm getting:
```
The decoder for Tiff does not support the format features Compression method Deflate is unsupported
```

Am I holding it wrong? Or is there some specific to this "Adobe Deflate" that ImageMagick uses that makes it different from Deflate supported by this crate?

RReverser:
- Switch to streaming decompression via `flate2`. Aside from performance improvements and lower RAM consumption, this fixes a bug where `max_uncompressed_length` was precalculated for a single tile but then used as a hard limit on the whole data, failing to decompress any tiled images.
 - Deprecate `InflateError` - this was used for wrapping `miniz_oxide` raw status, but `flate2` already reports them as `io::Error` instead. This makes errors from both Lzw and Deflate modes more consistent - they're all now just `io::Error`.
 - Add support for new `Deflate` tag in addition to `OldDeflate` - format-wise it's same Zlib, and on practice decodes successfully with the same underlying code.
 - Slightly simplify partial reads from the `Reader` by using the `std::io::copy` helper instead of doing it manually.

Fixes #131.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,10 +20,7 @@ edition = "2018"
 [dependencies]
 weezl = "0.1.0"
 jpeg = { package = "jpeg-decoder", version = "0.1.17", default-features = false }
-
-[dependencies.miniz_oxide]
-version = "0.4.1"
-features = ["no_extern_crate_alloc"]
+flate2 = "1.0.20"
 
 [dev-dependencies]
 criterion = "0.3.1"
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1084,7 +1084,7 @@ impl<R: Read + Seek> Decoder<R> {
         // Construct necessary reader to perform decompression.
         self.goto_offset_u64(offset)?;
         let byte_order = self.reader.byte_order;
-        let mut reader: Box<dyn Read> = match self.compression_method {
+        let reader: Box<dyn Read> = match self.compression_method {
             CompressionMethod::None => Box::new(&mut self.reader),
             CompressionMethod::LZW => Box::new(
                 LZWReader::new(
@@ -1097,8 +1097,8 @@ impl<R: Read + Seek> Decoder<R> {
             CompressionMethod::PackBits => {
                 Box::new(PackBitsReader::new(&mut self.reader, usize::try_from(length)?)?.1)
             }
-            CompressionMethod::OldDeflate => {
-                Box::new(DeflateReader::new(&mut self.reader, buffer.len() * buffer.byte_len())?.1)
+            CompressionMethod::Deflate | CompressionMethod::OldDeflate => {
+                Box::new(DeflateReader::new(&mut self.reader))
             }
             method => {
                 return Err(TiffError::UnsupportedError(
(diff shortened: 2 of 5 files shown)


fintelia:
This broadly looks good to me, but I'm slightly nervous about switching to flate2. There seems to be [outstanding soundness concerns](https://github.com/rust-lang/flate2-rs/issues/220) about the library. `image-rs` tends to take these sorts of things rather seriously both because image decoding can be security critical, and because there very high quality and well audited C/C++ image decoders that would be natural to use if people were comfortable with code that is "unsafe but probably fine."

I do really like the switch to `std::io::copy`, I hadn't known about that function.

Finally, before merging I would like to track down some documentation on the various deflate tags are why the different ones exist. Just to make sure we're not overlooking anything.

RReverser:
> There seems to be outstanding soundness concerns about the library.

IIRC, those are concerns if the code changes, right now it's working _fine_ - to the same extent as any wrapping around `unsafe` code works fine as long as some guarantees uphold. I agree switching to more robust wrappers like `Vec` or `MaybeUninit` would be even better on the `flate2` side though.

That is to say, I agree upstream code safety could be improved further, but my understanding is it's safe in its current state and I'm a bit worried that building similar streaming wrapper around `miniz_oxide` would only increase the risk of running into same bugs that have been already solved in `flate2` and tested by the rest of the ecosystem that relies on it directly.

> Finally, before merging I would like to track down some documentation on the various deflate tags are why the different ones exist.

Yeah, I'm not sure myself - that's what I tried to figure out first by looking up those different Deflate tags, but other than "it was specified in different documents / by different vendors" I personally couldn't find any meaningful explanation of why two exist. Then, I decided to just look at the extracted raw data and saw that it has the same Zlib magic, and decided to try and verify that it indeed works with the same Zlib code.

Note that thanks to Zlib using a magic header instead of just being raw DEFLATE stream, worst case scenario is, raw data won't match Zlib format and will still fail to parse even after this PR is merged - but on the images I've had and tested on so far it worked successfully.

fintelia (approved):
I just checked the `libtiff` source code, and they seem to handle the two deflate tags the same. We should be good on that front.

I'm also now noticing that flate2 is in the `rust-lang` organization which makes me feel much better about depending on it. Lets wait just a bit longer in case anyone else raises concerns, and otherwise I think this is ready to merge.

RReverser:
> I just checked the `libtiff` source code, and they seem to handle the two deflate tags the same.

Ah, good to know, thanks for checking!

RReverser:
> Lets wait just a bit longer in case anyone else raises concerns, and otherwise I think this is ready to merge.

@fintelia Do you think this is good to merge or did you mean to wait a bit longer? Once this part is released, I'd like to make a PR to the image crate to update TIFF dependency too (looks like it's a bit behind API-wise).

RReverser:
@fintelia Thanks! In terms of the release - should we wait for 0.7.0 as per https://github.com/image-rs/image-tiff/pull/130 (and, correspondingly, #129 as per https://github.com/image-rs/image-tiff/pull/130#issuecomment-824103275) or would this constitute a 0.6.2?

fintelia:
I don't think this warrants back-porting to 0.6.x, but we can hopefully get 0.7.0 out soon. I'll try to look over and resolve #129 this weekend.

RReverser:
Thanks, fair enough.

PR #128, opened 2021-04-11
Simplify expand_strip

fintelia:
This refactors the strip decoding logic with the goal of making it simpler and easier to modify.

PR #133, opened 2021-05-15
expand_strip() and expand_tile() cleanup

TomasKralCZ:
In this PR I attempted to make the expand_strip() and expand_tile() functions a little bit more dry. I also fixed inconsistencies between the 2 methods when dealing with color inversion for the signed formats.


tests/decode_images.rs
```diff
@@ -236,11 +236,6 @@ fn test_tiled_oversize_gray_i8() {
     test_image_sum_i8("tiled-oversize-gray-i8.tif", ColorType::Gray(8), 1214996);
 }
 
-#[test]
-fn test_tiled_whiteiszero_gray_i8() {
-    test_image_sum_i8("tiled-whiteiszero-gray-i8.tif", ColorType::Gray(8), 1214996);
-}
-
 #[test]
```
fintelia, line 239: Is there a reason this test is being removed?

src/decoder/mod.rs
```diff
@@ -1057,6 +1007,86 @@ impl<R: Read + Seek> Decoder<R> {
...
+    where
+        T: std::ops::Sub<T> + std::ops::Sub<Output = T> + Copy,
+    {
+        for datum in buffer.iter_mut() {
+            // FIXME: assumes [0, 1) range for floats
+            *datum = max - *datum
+        }
+    }
+
+    fn invert_colors(buf: &mut DecodingBuffer, color_type: ColorType, start: usize, end: usize) {
```
HeroicKatora, line 1029:
```suggestion
    use core::ops::Range;
    fn invert_colors(buf: &mut DecodingBuffer, color_type: ColorType, range: Range<usize>) {
```

src/decoder/mod.rs
```diff
@@ -1057,6 +1007,86 @@ impl<R: Read + Seek> Decoder<R> {
...
+            }
+            (ColorType::Gray(64), DecodingBuffer::F64(ref mut buffer)) => {
+                Self::invert_colors_fp(&mut buffer[start..end], 1.0);
+            }
+            _ => {}
+        }
+    }
+
+    /// Fix endianness. If `byte_order` matches the host, then conversion is a no-op.
+    fn fix_endianness(buf: &mut DecodingBuffer, byte_order: ByteOrder, s: usize, e: usize) {
```
HeroicKatora, line 1054: Same change here.

tests/decode_images.rs
TomasKralCZ: I thought we were going to remove the WhiteIsZero conversions for i8 and i16, so I removed the test as well. (Sorry if I misunderstood)

fintelia:
Instead of passing start and end arguments to `invert_colors` and `fix_endianness`, could you add a `DecodingBuffer::subrange()` method or similar modeled after the `DecodingBuffer::prefix()` method that already exists? In fact, since prefix is only used in a single place, it might even make sense to merge the two methods.

HeroicKatora (commented):
Using the `Range` type directly clarifies the signature quite a bit, I think. Also very nice consolidation of the duplicate logic here.

tests/decode_images.rs
TomasKralCZ: I read your last reply to the "signed conversions.." thread as "leave things as they are after https://github.com/image-rs/image-tiff/pull/128" but you probably meant "leave things as they are before https://github.com/image-rs/image-tiff/pull/128" ?

fintelia: Is that test failing now?

TomasKralCZ:
Thanks for the feedback, appreciate it :)

src/decoder/mod.rs
```diff
@@ -1128,77 +1143,12 @@ impl<R: Read + Seek> Decoder<R> {
...
-                DecodingBuffer::F32(buf) => buf
-                    .iter_mut()
-                    .for_each(|v| *v = f32::from_bits(u32::from_be(v.to_bits()))),
-                DecodingBuffer::F64(buf) => buf
-                    .iter_mut()
-                    .for_each(|v| *v = f64::from_bits(u64::from_be(v.to_bits()))),
-            },
-        }
+        let len = buffer.len();
+        Self::fix_endianness(&mut buffer.subrange(0..len), byte_order);
```
fintelia, line 1147: Should be no need for the `subrange` call here.

src/decoder/mod.rs
```diff
@@ -1128,77 +1143,12 @@ impl<R: Read + Seek> Decoder<R> {
...
-                }
-                (ColorType::Gray(64), DecodingBuffer::F64(ref mut buffer)) => {
-                    for datum in buffer.iter_mut() {
-                        // FIXME: assumes [0, 1) range for floats
-                        *datum = 1.0 - *datum
-                    }
-                }
-                _ => {}
-            }
+            Self::invert_colors(&mut buffer.subrange(0..len), color_type);
```
fintelia, line 1151: subrange call should be unnecessary here as well.

fintelia (approved):

Issue #89, opened 2020-08-31
Floating point predictor support


HeroicKatora:
Forked from #70 .

The horizontal predictor for floating points does not use float differences, as those would be lossy, but apparently uses an integer difference. Recheck this with the standard.

There is a dedicated floating point predictor which first re-orders some bytes: http://chriscox.org/TIFFTN3d1.pdf

TomasKralCZ:
I got the floating point predictor working, will try to polish the code a bit and submit a PR.

Here's a little visualization of how it works: (note that differencing is done _before_ reordering)
![visual](https://user-images.githubusercontent.com/23663355/118938225-b62e3000-b94e-11eb-9cb2-ef7e7973113e.png)

I only implemented the f32 version. Libtiff supports 16, 24, 32 and 64 bit versions. We currently don't support 16 and 24 bit floating-point samples at all, so I won't implement that but it won't be hard to add later. 

I guess I could add code for 64-bit samples as well, but I sadly don't have any test images in that format. All of my test images were generated using Photoshop, which doesn't support saving 64-bit samples.

PR #135, opened 2021-05-21
floating-point predictor added

TomasKralCZ:
This PR implements floaating-point predictor from https://github.com/image-rs/image-tiff/issues/89.

Only images with 32-bit samples are tested, as I don't have access to any 64-bit ones.


fintelia:
This looks good from a quick skim. Sorry, I haven't gotten a chance to read up on TIFF predictors and go through this in detail.

TomasKralCZ:
Oh, that's ok, this isn't urgent...

src/decoder/predictor.rs
```diff
@@ -0,0 +1,156 @@
...
+    for row in 0..height {
+        for col in samples..width * samples {
+            let prev_pixel = image[(row * img_width * samples + col - samples)];
+            let pixel = &mut image[(row * img_width * samples + col)];
+            *pixel = pixel.wrapping_add(prev_pixel);
+        }
+    }
+}
+
+pub fn fp_predict<T>(
```
fintelia, line 72: I'm not sure it makes sense for `fp_predict` to be generic. If there was just a `fp_predict_f32` and `fp_predict_f64` they could both be much simpler. I think you should be able to do something along the lines of: 
```rust
fn fp_predict_f32(buf: &mut [f32], (width, height): (usize, usize), res_width: usize, samples: usize) {
    for row in 0..height {
        rev_hpredict_nsamp(AsNeMutBytes::as_ne_mut_bytes(
            &mut buf[row * res_width * simples_per_pixel..][..width * samples_per_pixel]), ...);
        for i in 0..(width*samples_per_pixel) {
             buf[i] = f32::from_bits(buf[i].to_bits().from_be())
        }
    }
}

fintelia (commented):
Added some comments. I think the general structure could be significantly simplified, but the overall logic looks good to me.

src/decoder/predictor.rs
aschampion: FWIW there are 16- and 24-bit floating point TIFFs, the 16-bit version not being that rare in the wild so we might have an interest in supporting it at some point. Libtiff supports both.

TomasKralCZ:
Sorry, I was busy with other things, but I'll try to finish this soon...

fintelia:
https://github.com/image-rs/image-tiff/pull/165 is now merged. Thanks @TomasKralCZ for contributing the initial design!

PR #139, opened 2021-05-28
Check for various arithmetic overflows

5225225:
This causes a panic with multiplication overflow in debug mode (and, on this particular case, a `unreachable!` if overflow checking is disabled).

`LimitsExceeded` seems like the most reasonable error to emit, even though we don't care about the limits the user has set here. I could swap it out for `IntSizeError`, but it's not an error during *conversion*, it's an error during an *operation*.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1550,7 +1550,13 @@ impl<R: Read + Seek> Decoder<R> {
     }
 
     fn result_buffer(&self, width: usize, height: usize) -> TiffResult<DecodingResult> {
-        let buffer_size = width * height * self.bits_per_sample.len();
+        let buffer_size = match width
+            .checked_mul(height)
+            .and_then(|x| x.checked_mul(self.bits_per_sample.len()))
+        {
+            Some(s) => s,
+            None => return Err(TiffError::LimitsExceeded),
+        };
 
         let max_sample_bits = self.bits_per_sample.iter().cloned().max().unwrap_or(8);
         match self.sample_format.first().unwrap_or(&SampleFormat::Uint) {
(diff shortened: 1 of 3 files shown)


5225225:
I'll keep looking for fairly trivial fixes and move them into this pull request. There is an OOM where the limits aren't being respected in the decompressor that I'm not as sure about how to fix, so I'll split that out.

src/decoder/mod.rs
```diff
@@ -1277,7 +1277,13 @@ impl<R: Read + Seek> Decoder<R> {
             return Ok(0);
         }
 
-        Ok((self.height + rows_per_strip - 1) / rows_per_strip)
+        // rows_per_strip - 1 can never fail since we know it's at least 1
+        let height = match self.height.checked_add(rows_per_strip - 1) {
+            Some(h) => h,
+            None => return Err(TiffError::IntSizeError),
```
5225225, line 1283: I'm not entirely sure if IntSizeError is the right error for here. Probably?

The places where I used LimitsExceeded is when, if the operation was on a bigint and worked, we'd try to read a huge amount of data, so I'm treating it as if the user always has a limit of `usize::MAX`.

HeroicKatora: An `IntSizeError` seems to be appropriate.

5225225:
Okay, ran the fuzzer for a few minutes and got a lot of (probably mostly the same) OOMs, but no crashes, so I feel the best thing here is to get these in, then fix the OOMs, then fix any other issues deeper in the library that couldn't be found due to the shallower bugs. (And also add the fuzzer itself to the repo).

Going to clean up the commits (just making sure tests pass on each one and fixing the rustfmt on e5b8af7), then this should be good to review.

HeroicKatora (approved):
Sounds good. When adding the fuzzer you can also add the regression tests as seeds in a separate directory. If you have any questions about that we can discuss this in the fuzzer PR.

PR #140, opened 2021-05-29
Add fuzzer with 1MB size limits decoding image

5225225:
re: https://github.com/image-rs/image-tiff/pull/139#pullrequestreview-671738717

By adding the regression tests in a separate directory, do you mean `tests/fuzz_images/<category>`?

I've not really been saving the files for the failures, since I figure it's easy enough to find more failing test cases as needed. I can start saving them in that director. I've just been adding the first one I find of a particular failure to the test `.rs` file as an array.

And/or should I save the corpus somewhere?

### DIFF
--- a/fuzz/fuzz_targets/decode_image.rs
+++ b/fuzz/fuzz_targets/decode_image.rs
@@ -14,6 +14,14 @@ fuzz_target!(|data: &[u8]| {
     limits.intermediate_buffer_size = 1_000_000;
 
     decoder = decoder.with_limits(limits);
+    
+    loop {
+        if let Err(_) = decoder.read_image() {
+            break;
+        }
 
-    let _ = decoder.read_image();
+        if !decoder.more_images() {
+            break;
+        }
+    }
 });
--- a/fuzz/fuzz_targets/decode_image.rs
+++ b/fuzz/fuzz_targets/decode_image.rs
@@ -1,0 +1,19 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    let mut decoder = if let Ok(d) = tiff::decoder::Decoder::new(std::io::Cursor::new(data)) {
+        d
+    } else {
+        return;
+    };
+
+    let mut limits = tiff::decoder::Limits::default();
+    limits.decoding_buffer_size = 1_000_000;
+    limits.ifd_value_size = 1_000_000;
+    limits.intermediate_buffer_size = 1_000_000;
+
+    decoder = decoder.with_limits(limits);
+
+    let _ = decoder.read_image();
+});
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,3 @@
+target
+corpus
+artifacts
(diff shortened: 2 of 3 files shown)


HeroicKatora:
> By adding the regression tests in a separate directory, do you mean tests/fuzz_images/<category>?

Yes, that's what I meant. At least if you have manually classified regression tests, that is. However for a full automatically generated (and minified) corpus we might add the `fuzz/corpus` directory itself. At least if it is decently small and somewhat complete this seed directory is valuable. Maybe we should spend about a cpu month on creating such a seed directory.

5225225:
> At least if you have manually classified regression tests, that is.

I don't at the moment, no, I generally just delete the crashing inputs after I've written the test for that crash. My plan was to add test cases in the `.rs` file for the OOMs as I find them (as I've been doing), but to otherwise not save them.

Generating a corpus might be difficult as is, but once I fix the OOMs and can let the fuzzer run overnight, I can add that in a future pull request.

I'd want to do it in a way that means git's not trying to track the files as you run the fuzzer, since it's annoying to run the fuzzer and then have hundreds of changed files, but I can sort out how to do that then.

HeroicKatora:
You don't need to worry about the fuzzer runs, when we're at clean enough state with few crashes i can try and organize a weekend run on a threadripper machine. That should be good enough for a start :slightly_smiling_face: 

Regarding git tracking files, as long as the corpus directory itself is not added there shouldn't be any problem. I don't think they any of the corpus files get changed or even deleted by default when running the fuzzer. In other words, one could force-add (`git add -f`) a somewhat complete corpus even with the corpus directory being ignored usually. I'm not sure about the priority of being able to minimize the corpus but maybe you can try it this way.

5225225:
This good to be merged as-is, or are there any other things that should be changed? I created #141 as a draft because it contains this PR as well since I needed the fuzzer to find the bugs, so I can either close this PR and have the fuzzer and the bugfixes reviewed at the same time, or merge this one then I'll open #141 for review.

I left the fuzzer running on all cores for an hour or so on #141 and couldn't find anything else, so there's probably no more easy to find bugs.

fuzz/fuzz_targets/decode_image.rs
```diff
@@ -0,0 +1,19 @@
...
+    };
+
+    let mut limits = tiff::decoder::Limits::default();
+    limits.decoding_buffer_size = 1_000_000;
+    limits.ifd_value_size = 1_000_000;
+    limits.intermediate_buffer_size = 1_000_000;
+
+    decoder = decoder.with_limits(limits);
+
+    let _ = decoder.read_image();
```
HeroicKatora, line 18: This decodes exactly one image which might miss issues that appear when multiple images appear consecutively within the same file, or non-termination due to missing the end of file etc.. It would be proper to wrap this in a while-loop checking `reader.more_images()` and calling `reader.next_image` or simply iterating like so

> `while let Ok(_) = reader.next_image()`

HeroicKatora (commented):
All the fuzzer setup and ignore looks good and the limits look reasonable for fuzzing as well.

fuzz/fuzz_targets/decode_image.rs
5225225: Pushed what I *think* is a correct implementation.

And yeah, you're right, I can find timeouts and overflows with that change. So looks like I have more fixes to do!

HeroicKatora (approved):

Issue #137, opened 2021-05-27
Decoder regression: mandrill.tiff from upstream


RReverser:
I've tried to upgrade upstream `image` crate to tiff 0.7.0, and it uncovered a regression bug: upstream image [mandrill.tiff](https://github.com/image-rs/image/blob/master/tests/images/tiff/testsuite/mandrill.tiff) used to decode with 0.6.0, but fails with 0.7.0 with:

> Cannot create decoder: FormatError(Format("Neither strips nor tiles were found or both were used in the same file"))

I verified that this problem is reproducible with just `image-tiff` too, so it's not an issue with integration in the `image` crate itself. It's also an uncompressed TIFF, so doesn't seem related to my deflate PR (#132) either.

Beyond that, I'm not that familiar with changes between releases, so leaving it up to others to debug this further.

TomasKralCZ:
Sorry, this is the result of some code I added.
https://github.com/image-rs/image-tiff/blob/bd1e37d78b731d718200e8909a89e53c93c4df4a/src/decoder/mod.rs#L704-L719

This file has StripOffsets and StripByteCounts, but doesn't have RowsPerStrip, which I didn't expect.
Maybe I should just check for (StripOffsets & StripByteCounts) | (TileOffsets & TileByteCounts), otherwise reject the file ?

TomasKralCZ:
Sorry for taking so long, I was busy with studying.
Fixing this isn't as simple as I thought. I'll make a PR soon.

PR #145, opened 2021-06-16
Fix the mandrill.tiff regression

TomasKralCZ:
This fixes https://github.com/image-rs/image-tiff/issues/137.
I'd consider this to be more of a temporary workaround, hence the TODO.
ChunkType is related to incremental decoding, so fixing this in a better way would mean changing the way strip incremental decoding works. So that should probably be done in a seperate PR.

The `test_div_zero()` was removed because it was now (properly) returning an error in this function.


HeroicKatora:
> The test_div_zero() was removed because it was now (properly) returning an error in this function.

I'm not sure if I understand this. This was introduced as a regression test for a certain type of fuzzer failure (https://github.com/image-rs/image-tiff/pull/138) so rather I would update it to ignore the error. The main concern of the test is with panicking.

TomasKralCZ:
Oh, ok, I'll update the error.

src/decoder/mod.rs
```diff
@@ -699,20 +701,20 @@ impl<R: Read + Seek> Decoder<R> {
...
         self.chunk_type =
+            // TODO: This needlessly allocates a vector and then discards it. Maybe init decoders here.
             match (
-                self.get_tag_u32(Tag::RowsPerStrip),
-                self.get_tag_u32(Tag::TileWidth),
-                self.get_tag_u32(Tag::TileLength),
+                self.get_tag_u64_vec(Tag::StripByteCounts),
+                self.get_tag_u64_vec(Tag::StripOffsets),
+                self.get_tag_u64_vec(Tag::TileByteCounts),
+                self.get_tag_u64_vec(Tag::TileOffsets),
```
fintelia, line 709: Couldn't you just do `self.ifd.as_ref().unwrap().contains_key(&tag)` instead for each of these? That would avoid the read and the vector allocation.

TomasKralCZ: Thank  you, that's  much better... I feel kinda dumb, because I spent some time looking for something like contains_key(), but didn't realize that Directory is a HashMap.

HeroicKatora (approved):
LGTM

fintelia (approved):

RReverser:
It's been a while, can a release be tagged please?

PR #159, opened 2022-04-15
Make LZWReader incrementally decompress input

fintelia:
Avoids a potentially large allocation.


tests/decode_images.rs
```diff
@@ -409,7 +409,7 @@ fn oom_2() {
     let err = decoder.read_image().unwrap_err();
 
     match err {
-        tiff::TiffError::LimitsExceeded => {}
+        tiff::TiffError::IoError(_) => {}
         unexpected => panic!("Unexpected error {}", unexpected),
```
HeroicKatora, line 413: It's no longer accurate to call this `oom_2` now.

src/decoder/stream.rs
```diff
@@ -131,63 +131,94 @@ pub type DeflateReader<'r, R> = flate2::read::ZlibDecoder<&'r mut SmartReader<R>
...
 
 /// Reader that decompresses LZW streams
-pub struct LZWReader {
-    buffer: io::Cursor<Vec<u8>>,
+pub struct LZWReader<R: Read> {
+    compressed_buf: Box<[u8]>,
+    compressed_buf_length: usize,
+    compressed_bytes_read: usize,
+
+    reader: R,
```
HeroicKatora, line 139: Isn't that more or less a duplicate of `BufReader<Take<R>>`?

src/decoder/stream.rs
```diff
@@ -131,63 +131,94 @@ pub type DeflateReader<'r, R> = flate2::read::ZlibDecoder<&'r mut SmartReader<R>
...
+        self.compressed_bytes_read += result.consumed_in;
+        self.output_remaining -= result.consumed_out;
+
+        match result.status {
+            Ok(weezl::LzwStatus::Ok) => {
+                if result.consumed_out == 0 {
+                    return self.read(buf);
+                }
+            }
+            Ok(weezl::LzwStatus::NoProgress) => {
```
HeroicKatora, line 194: `NoProgress` indicates that no bytes from the input were consumed, in addition to the condition recognized in the assert_eq. Input bytes will be consumed if a new symbol must be read. So: I'd rather we check that `buf` is non-empty at the start of the loop by returning `Ok(0)` if it is.

Then, `BufReader::fill_buf` makes sure the input is not empty in all cases except eof. Since `NoProgress` becomes equivalent 'empty input stream' when the output buffer is non-empty, we have reached EOF here.

src/decoder/stream.rs
fintelia: The frustrating thing is that `BufReader` has no "grow buffer" functionality. `BufReader::fill_buf` only guarantees that the internal buffer will have at least one byte in the internal buffer (assuming there is more input), but weezl may need multiple bytes of input to successfully advance the internal state.

HeroicKatora: Hm? That doesn't sound right. As long as there's an output buffer and a non-empty code buffer it should advance state _somehow_. It also does [partial refills](https://github.com/image-rs/lzw/blob/master/src/decode.rs#L997-L1002) of the internal codeword bit buffer if the input is too short for a full word.

HeroicKatora: If you have a test to the contrary–that is lack of progress with a non-empty input **and** output buffer– I'd regard that as a bug to `weezl`.

fintelia: The documentation of LzwStatus::NoProgress says "No bytes were read or written and no internal state advanced."

fintelia: But looking in more detail, it seems the tests that were triggering NoProgress were doing so by passing an empty input buffer

HeroicKatora: There is no minimum input length to advance though. As long as you provide bytes and decoding isn't stuck on writing, it will advance input state and consume bytes. There's no reason that requires filling the input buffer to a _any_ minimum size.

src/decoder/stream.rs
fintelia: Realized that calls to this reader are always wrapped in io::Take<> so we don't actually need to ever worry about returning Ok(0). If the caller wants more input and there's none left, that is always an error

HeroicKatora (approved):
LGTM with a rustfmt run

PR #160, opened 2022-04-15
Make PackBitsReader incrementally decompress output

fintelia:
Same idea as #159

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1141,9 +1141,10 @@ impl<R: Read + Seek> Decoder<R> {
             CompressionMethod::LZW => {
                 Box::new(LZWReader::new(reader, usize::try_from(compressed_length)?))
             }
-            CompressionMethod::PackBits => {
-                Box::new(PackBitsReader::new(reader, usize::try_from(compressed_length)?)?.1)
-            }
+            CompressionMethod::PackBits => Box::new(PackBitsReader::new(
+                reader,
+                usize::try_from(compressed_length)?,
+            )),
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -1141,9 +1141,10 @@ impl<R: Read + Seek> Decoder<R> {
             CompressionMethod::LZW => {
                 Box::new(LZWReader::new(reader, usize::try_from(compressed_length)?))
             }
-            CompressionMethod::PackBits => {
-                Box::new(PackBitsReader::new(reader, usize::try_from(compressed_length)?)?.1)
-            }
+            CompressionMethod::PackBits => Box::new(PackBitsReader::new(
+                reader,
+                usize::try_from(compressed_length)?,
```
HeroicKatora, line 1146: This `usize::try_from` can be removed if `new` method is changed to accept `u64` to begin with.

src/decoder/stream.rs
```diff
@@ -255,56 +255,70 @@ impl Read for JpegReader {
...
+            }
+            PackBitsReaderState::Header => unreachable!(),
+        }
 
-impl Read for PackBitsReader {
-    #[inline]
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        self.buffer.read(buf)
+        self.count -= length;
+        if self.count <= buf.len() {
```
HeroicKatora, line 318: This condition seems rather confusing to me. Shouldn't it be `self.count == 0` given that we consume the number of written bytes in the previous line?

HeroicKatora (approved):

PR #161, opened 2022-04-16
Parallel decoding

fintelia:
This PR adds parallel decoding via an optional (but default) dependency on rayon.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1221,9 +1221,21 @@ impl<R: Read + Seek> Decoder<R> {
         {
             let condvar = std::sync::Condvar::new();
             let bytes_allocated = std::sync::Mutex::new(Ok(0));
+
+            // Loop over columns of chunks (for stripped images `chunks_across` is always 1).
+            //
+            // This iteration is needed to work around an annoying lifetime issue for tiled images:
+            // each tile of an image will map to potentially thousands of non-contiguous ranges in
+            // the output buffer. So instead of getting mutable references to a single tile, we
+            // take a mutable refence to a entire rows of tiles (which do fit within a single
+            // contiguous range) and on each iteration only access the x-th tile in that row.
             for x in 0..chunks_across {
                 rayon::in_place_scope_fifo(|s| -> TiffResult<()> {
                     let buffer_offset = x * chunk_dimensions.0 * samples;
+                    // Iterate over rows of chunks. For stripped images, `strip` will be the y-th
+                    // strip of the image. While for tiled images, `strip` will be the memory for
+                    // the entire y-th row of tiles so the loop body will be careful to only access
+                    // the parts corresponding to tile (x,y).
                     for (y, strip) in result
                         .as_buffer(buffer_offset)
                         .chunks_mut(strip_samples)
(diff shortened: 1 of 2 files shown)


src/decoder/mod.rs
```diff
@@ -1455,60 +1218,159 @@ impl<R: Read + Seek> Decoder<R> {
...
+                .image()
+                .chunk_bytes
+                .iter()
+                .all(|&b| b <= self.limits.intermediate_buffer_size as u64 / 2)
         {
-            Some(s) => s,
-            None => return Err(TiffError::LimitsExceeded),
-        };
+            let condvar = std::sync::Condvar::new();
+            let bytes_allocated = std::sync::Mutex::new(Ok(0));
```
HeroicKatora, line 1278: Somewhat questionable to me? Blocking a rayon worker seems very unintuitive, and after the jpeg-decoder experience there should be more comments on correctness/forward progress.

src/decoder/mod.rs
```diff
@@ -1455,60 +1218,159 @@ impl<R: Read + Seek> Decoder<R> {
...
+            for x in 0..chunks_across {
+                rayon::in_place_scope_fifo(|s| -> TiffResult<()> {
```
HeroicKatora, line 1280: We're already starting a new scope for each x coordinate, can't we ensure the allocation requirement by ensuring that at most `n` tasks run in parallel, where `n` is `intermediate_buffer_size / chunk_bytes.iter().max()`. Or an adaptive version instead.

I'd also be more comfortable if tasks were generated as an iterator instead of this somewhat complex case distinction of different loops, if that makes sense. In particular this makes it easier to transition to using some of the rayon parallel iterators in the future?

src/decoder/image.rs
```diff
@@ -287,4 +268,196 @@ impl Image {
...
+            }
+            method => {
+                return Err(TiffError::UnsupportedError(
+                    TiffUnsupportedError::UnsupportedCompressionMethod(method),
+                ))
+            }
+        })
+    }
+
+    pub(crate) fn chunk_byte_range(&self, chunk: usize) -> TiffResult<(u64, u64)> {
```
HeroicKatora, line 336: Can we rename this to `chunk_file_range`, or similar, to better differentiate it from `chunks_mut` et.al that operate on in-memory buffers?

HeroicKatora (commented):
I'm not sure if the task strategy is correct yet. It doesn't seem very 'rayon idiomatic'.

I'd be more comfortable with the task strategy if we allocate a fixed number of parallel resources (that is, prepare buffers for some determined number of parallelism that depends on the required buffer sizes vs. available allocation size vs. rayon's available core count). And have an iterator defining the chunks+target buffers that need processing.

If those task-resources are in a `vec<_>` then we can use [`flat_map_iter`](https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.flat_map_iter) to actually execute the task. Each task could then create an 'iterator' that polls from the shared chunk iterator (by some means of synchronization). That would guarantee a precise upper bounds on the allocated buffer size while also preserving those allocations between chunks.

Can you imagine something like this working?

src/decoder/mod.rs
fintelia: Added a bunch of comments. The general idea is that taking the lock should never deadlock because all threads are careful to never do any blocking operations while holding the lock. We do sleep on the condition variable from the calling threads, but never from one of rayon's worker threads. As long as the worker threads make forward progress the calling thread should eventually be woken up.

src/decoder/mod.rs
fintelia: The two loops are needed to work around a borrow checker issue for tiled images (for stripped images the outer loop executes exactly one iteration).

It may be possible to replace the inner loop with an iterator that generate the tasks. However, I think it would be tricky to implement that while still doing file reads and chunk expands in parallel

fintelia:
@HeroicKatora how do you feel about the current state? If you'd prefer we iterate more, I can split out the rayon parallelized expand_chunk function into a separate PR to let us merged the the other code movement/refactoring parts. That would reduce potential for merge conflicts with other concurrent PRs

HeroicKatora:
Sounds good to split it up. Consolidating chunk is a very good and orthogonal change compared to the prior explicit line/tile differentiation.

fintelia:
Ok, I've rebased this on the main branch so now this just includes the parallel decoding part.

I also into other ways of using rayon for decoding. Unfortunately, we're rather constrained by the lack of a `Send` bound on the reader. The `in_place_scope[_fifo]` methods are pretty much the only option. Even [`ParallelBridge`](https://rust-lang.github.io/hashbrown/rayon/iter/trait.ParallelBridge.html) requires a `Send`-able iterator.

PR #162, opened 2022-04-27
Added ability to decompress tiled jpeg files

AlanRace:
Currently compression type 7 (`ModernJPEG`) is not supported when expanding either a strip or a tile. This means that it is not possible to load a single tile of an image (e.g. via `read_tile`) with this compression (only `read_image` can be used). 

To make sure that the correct data is displayed for the test data I was using, I had to incorporate the changes from https://github.com/image-rs/image-tiff/pull/148 too.

This also removes the special case for handling `ModernJPEG` in `read_image`.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1533,9 +1533,9 @@ impl<R: Read + Seek> Decoder<R> {
 
     /// Decodes the entire image and return it as a Vector
     pub fn read_image(&mut self) -> TiffResult<DecodingResult> {
-        let result = match (self.image().chunk_type, self.image().compression_method) {
-            (ChunkType::Strip, _) => self.read_stripped_image()?,
-            (ChunkType::Tile, _) => self.read_tiled_image()?,
+        let result = match self.image().chunk_type {
+            ChunkType::Strip => self.read_stripped_image()?,
+            ChunkType::Tile => self.read_tiled_image()?,
         };
 
         Ok(result)
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::convert::TryFrom;
 use std::io::{self, Cursor, Read, Seek};
+use std::sync::Arc;
 use std::{cmp, ops::Range};
 
 use crate::{
@@ -1151,7 +1152,7 @@ impl<R: Read + Seek> Decoder<R> {
         photometric_interpretation: PhotometricInterpretation,
         compression_method: CompressionMethod,
         compressed_length: u64,
-        jpeg_tables: Option<Vec<u8>>,
+        jpeg_tables: Option<Arc<Vec<u8>>>,
     ) -> TiffResult<Box<dyn Read + 'r>> {
         Ok(match compression_method {
             CompressionMethod::None => Box::new(reader),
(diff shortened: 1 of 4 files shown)


src/decoder/mod.rs
```diff
@@ -1132,18 +1129,43 @@ impl<R: Read + Seek> Decoder<R> {
     }
 
     fn create_reader<'r>(
-        reader: &'r mut SmartReader<R>,
-        compression_method: CompressionMethod,
+        decoder: &'r mut Self,
```
fintelia, line 1132: Taking a mutable reference to the entire reader conflicts with [ongoing work to enable multiple strips/tiles to be decoded in parallel](https://github.com/image-rs/image-tiff/pull/161). Would you instead be able to pass _jpeg_tables_ and _photometric_interpretation_ in as additional arguments? To avoid cloning the entire Vec, the type on _Image::jpeg_tables_ could be changed to `Option<Arc<Vec<u8>>>`

tests/decode_images.rs
```diff
@@ -391,6 +391,10 @@ fn oom() {
 
     match err {
         tiff::TiffError::LimitsExceeded => {}
+        tiff::TiffError::UnsupportedError(tiff::TiffUnsupportedError::InterpretationWithBits(
+            _,
+            _,
+        )) => {}
```
fintelia, line 397: nit: this can be shortened to `tiff::TiffError::UnsupportedError(tiff::TiffUnsupportedError::InterpretationWithBits(..)) => {}`

tests/decode_images.rs
```diff
@@ -391,6 +391,10 @@ fn oom() {
 
     match err {
         tiff::TiffError::LimitsExceeded => {}
```
fintelia, line 393: This line should be removed if the error below is now the one our decoder returns for this fuzzer generated test case.

fintelia (changes requested):
Our ModernJPEG decoding is severely under-tested. Would it be possible for you to share the test data you're using and/or find some other small freely available images to use as test cases?

src/decoder/mod.rs
AlanRace: > To avoid cloning the entire Vec, the type on Image::jpeg_tables could be changed to Option<Arc<Vec<u8>>>

I thought about this, but having looked at the code in `JpegReader::new()` a clone was also required as the `jpeg_tables` is then extended with the data. I thought the simplest way to reduce the number of clones was to require an owned Vec to be passed to `JpegReader::new()`. I'll commit a version with this, and you can decide if I should still change over to `Option<Arc<Vec<u8>>>`.

AlanRace:
> Would it be possible for you to share the test data you're using and/or find some other small freely available images to use as test cases?

I don't have any small files unfortunately, but there are many in the same style/format (.svs) publicly available on [https://portal.gdc.cancer.gov/](https://portal.gdc.cancer.gov/repository?filters=%7B%22op%22%3A%22and%22%2C%22content%22%3A%5B%7B%22op%22%3A%22in%22%2C%22content%22%3A%7B%22field%22%3A%22files.data_format%22%2C%22value%22%3A%5B%22svs%22%5D%7D%7D%5D%7D)

This format is essentially a tiff file with additional/custom tags. As they are whole (microscopy) slide images, they are typically > 300 MB per file.

src/decoder/mod.rs
```diff
@@ -1505,10 +1546,9 @@ impl<R: Read + Seek> Decoder<R> {
 
     /// Decodes the entire image and return it as a Vector
     pub fn read_image(&mut self) -> TiffResult<DecodingResult> {
-        let result = match (self.image().chunk_type, self.image().compression_method) {
-            (_, CompressionMethod::ModernJPEG) => self.read_jpeg()?,
-            (ChunkType::Strip, _) => self.read_stripped_image()?,
-            (ChunkType::Tile, _) => self.read_tiled_image()?,
+        let result = match self.image().chunk_type {
```
fintelia, line 1549: nit: instead of having a result variable, we could just wrap the entire match statement in `Ok(...)`

src/decoder/mod.rs
```diff
@@ -1505,10 +1546,9 @@ impl<R: Read + Seek> Decoder<R> {
 
     /// Decodes the entire image and return it as a Vector
     pub fn read_image(&mut self) -> TiffResult<DecodingResult> {
-        let result = match (self.image().chunk_type, self.image().compression_method) {
-            (_, CompressionMethod::ModernJPEG) => self.read_jpeg()?,
```
fintelia, line 1509: Now that read_stripped_image / read_tiled_image work with jpeg images, could you replace the contents with something like this (perhaps with a check that the image actually uses jpeg compression):
```rust
    #[deprecated = "Use read_image instead"]
    pub fn read_jpeg(&mut self) -> TiffResult<DecodingResult> {
        self.read_image()
    }
```

src/decoder/stream.rs
```diff
@@ -244,6 +252,58 @@ impl JpegReader {
...
 }
 
+#[repr(u8)]
+enum JpegTagApp14Transform {
+    // App14TransformYCCK = 2,
+    // App14TransformYCbCr = 1,
+    App14TransformUnknown = 0,
+}
+
+fn add_app14segment(jpeg_tables: &mut Vec<u8>, transform: JpegTagApp14Transform) {
```
fintelia, line 262: I'm not especially confident that this code is correct since I don't know much about JPEG encoding and we don't have any testcases that exercise it. Which is probably related to why the previous PR lingered. But it does fix decoding for some images. So, even though I'm not thrilled about it, I still think we should probably just merge this now.

src/decoder/mod.rs
```diff
@@ -1089,11 +1098,17 @@ impl<R: Read + Seek> Decoder<R> {
         self.goto_offset_u64(offset)?;
 
         let line_samples = output_width * self.image().bits_per_sample.len();
-        let compression_method = self.image().compression_method;
         let photometric_interpretation = self.image().photometric_interpretation;
+        let compression_method = self.image().compression_method;
+        let jpeg_tables = self.image().jpeg_tables.clone();
```
fintelia, line 1103: This could be a potentially large allocation, right? Could we use an Arc to avoid that?

AlanRace: I have now changed it to pass an Arc<Vec<u8>> instead

src/decoder/stream.rs
AlanRace: I guess one possibility would to alter `jpeg-decoder::Decoder` to allow overriding the `color_transform`? 

I can have a look at the `jpeg_tables` for the styles of data that I have, to see whether I can see where the colour is defined/why the original code treated it as YCrCb instead of RGB, but I think any change to try and resolve this in a more satisfactory manner would constitute a separate PR.

fintelia (approved):

PR #163, opened 2022-05-02
Add ability to select image directory and tile using index

AlanRace:
Currently it is only possible to cycle through the image directories and the tiles in order. For handling WSI files (e.g. .svs), where the image data is stored as a 'pyramid' (each image directory has a copy of the image at a different resolution), I need to be able to select which tiles (part of an image) and which image directory (which resolution) I want to access. 

This PR adds in this functionality, and can be used as so:

```rust
// Select the image directory at the third position (4th directory) in the linked list
decoder.image_at(3)?;

let tiles = decoder.tile_count()?;
let (width, height) = decoder.dimensions()?;

println!("# tiles {}", tiles);

for tile in 0..tiles {
    // As some tiles (on the boundaries) do not have the same size 
    // as the rest (due to padding) we need to be able to determine its size
    let chunk_info = decoder.chunk_info(tile as usize)?;

    match decoder.read_tile_at(tile as usize)? {
            DecodingResult::U8(res) => {
                // Do some thing with the tile, e.g. save out as a .png 
                image::save_buffer(
                    format!("image_{}.png", tile),
                    &res,
                    chunk_info.data_width as u32,
                    chunk_info.data_height as u32,
                    image::ColorType::Rgb8,
                )?;
            }
            _ => panic!("Wrong bit depth"),
    }
}
```


HeroicKatora (changes requested):
The capability would be very nice to have, I have a few questsiont regarding interface and implementation though.

src/decoder/mod.rs
```diff
@@ -1071,28 +1151,67 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(())
     }
 
+    pub fn chunk_info(&self, chunk_index: usize) -> Option<ChunkInfo> {
```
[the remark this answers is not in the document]
AlanRace: I have now removed public and am now returning ChunkInfo from `read_tile()` and `read_tile_at()`. To keep consistency I have added in the same functionality for accessing strips - but I believe this will be consolidated anyway in a different PR?

src/decoder/mod.rs
```diff
@@ -585,17 +625,58 @@ impl<R: Read + Seek> Decoder<R> {
         &self.image
     }
 
+    /// Loads the IFD at the specified index in the list, if one exists
+    pub fn image_at(&mut self, ifd_index: usize) -> TiffResult<()> {
```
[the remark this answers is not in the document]
AlanRace: How is `seek_to_image`?

src/decoder/mod.rs
```diff
@@ -1390,43 +1478,44 @@ impl<R: Read + Seek> Decoder<R> {
     }
 
     /// Read a single strip from the image and return it as a Vector
-    pub fn read_strip(&mut self) -> TiffResult<DecodingResult> {
-        self.check_chunk_type(ChunkType::Strip)?;
-        let index = self.current_chunk;
+    pub fn read_strip(&mut self) -> TiffResult<(DecodingResult, ChunkInfo)> {
```
HeroicKatora, line 1481: The comment regarding `ChunkInfo` was only referring to new interfaces, this would be a breaking change. Please revert the interface change here and in `read_tile`, and only in those two instances. 

This is a loss of information because it drops information in the case of the last strip (which need not contain the full `rows_per_strip` rows but might be less) and border tiles. This is slightly unfortunate, so adding a comment that this interface could be changed in a future release would be nice.

HeroicKatora (commented):
I should have been clearer regarding the comments regarding the interface, a breaking change to existing methods wasn't the intention.

Should be easy to fix, just drop one of the result from the `read_*_at` call in those two methods pre-existing methods. This is still somewhat consistent:

```
pub fn read_tile(&mut self) -> TiffResult<DecodingResult>
pub fn read_strip(&mut self) -> TiffResult<DecodingResult>

pub fn read_tile_at(&mut self) -> TiffResult<(DecodingResult, ChunkInfo)>
pub fn read_strip:at(&mut self) -> TiffResult<(DecodingResult, ChunkInfo)>
```

Some more documentation for new interfaces would also be nice to have.

src/decoder/mod.rs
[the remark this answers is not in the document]
HeroicKatora: Sounds fine :+1:

fintelia:
There's some smaller things I see, but the first thing to decide on is what the API of this should be. Some thoughts:

* Once our parallel decoding lands, in many cases decoding a full image is going to be much faster than doing the equivalent `read_strip` or `read_tile` calls one by one. It feels unfortunate to add another public API that also cannot leverage parallelism.

* If the caller already has destination buffer they'd like to use, doing in-place decoding (like `read_strip_to_buffer`) is going to avoid an unnecessary copy.

* It might be better to have a single `read_chunk_at` instead of one function for stripped images and another for tiled images.

* Instead of `seek_to_image`, perhaps we should pass the image index as a function argument to `read_*_at`?

HeroicKatora:
> If the caller already has destination buffer they'd like to use, doing in-place decoding (like read_strip_to_buffer) is going to avoid an unnecessary copy.

That'd be something that needs to be a addressed for all interfaces, regardless. Perhaps as `0.8`? Because it seems this change would be similar in scope as image's transition in `ImageDecoder`.

AlanRace:
My thoughts on your points...

 > *   Once our parallel decoding lands, in many cases decoding a full image is going to be much faster than doing the equivalent read_strip or read_tile calls one by one. It feels unfortunate to add another public API that also cannot leverage 
parallelism.

I can see three use cases here:
1. I want to load the whole image - this is perfectly well handled by the current public API so no change needed.
2. I want to load a region of the image (this is often the case in WSI as the whole image is very large, but I might only be viewing a small region at high resolution) - this is my primary use case for wanting to be able to load per tile and as you said with the proposed API would lose out on parallelisation. We could consider adding in something like `read_subimage(x: usize, y: size, width: usize, height: usize)` which could then benefit from internal parallelisation?
3. I want to process each tile separately (this can be the case in processing pathology images - e.g. I want to preprocess and then classify on a per-tile basis) - in this situation it doesn't make sense to load in the whole image (or part of it) using internal parallelisation and then split it back up to process per-tile. I would ideally like to handle the parallelisation myself, outside of `image-tiff`.

> *  It might be better to have a single read_chunk_at instead of one function for stripped images and another for tiled images.

I am happy to do this, but I didn't want to replicate some of your work (i.e. in https://github.com/image-rs/image-tiff/pull/161) for unifying tile/strip handling.

  > * Instead of seek_to_image, perhaps we should pass the image index as a function argument to read_*_at?

I am against this suggestion if the internal handling remains as it is. If it were changed as suggested, then I would expect as a user that I could do something like:
```rust
decoder.read_chunk_at(image1, chunk5);
decoder.read_chunk_at(image2, chunk5);
decoder.read_chunk_at(image1, chunk6);
```
with very little overhead, but in reality it will (possibly) parse the IFD for `image1`, then definitely parse the IFD for `image2`, then reparse the IFD for `image1` again. I think with the current implementation, `seek_to_image` makes it clearer the cost of calling this function.

Having said that, I would actually prefer to be able to completely randomly access tiles from different images without such an overhead - but that would likely require storing all IFDs in memory to avoid needing to reparse each time. What do you think?

src/decoder/mod.rs
AlanRace: Yeah I should have realised that is what you meant - I was surprised at the suggested change.

HeroicKatora (approved):
All reasoning sounds consistent to me, I tend to agree with all points being made. Maybe we could turn all three use-cases into public examples?

> Having said that, I would actually prefer to be able to completely randomly access tiles from different images without such an overhead - but that would likely require storing all IFDs in memory to avoid needing to reparse each time. What do you think?

Right, that's something better left for another day. One option that comes to mind to make this more orthogonal is defining a separate `IfdCache` struct, then providing an adaptor that borrows from both a cache and reader to execute those random-access operations. (For example, having `IfdCache::read_with(&mut decoder) -> CacheReader<'_>` or reverse argument order). That avoids consuming more memory in the decoder from having to store all the directories, the cache can be pre-filled (and persisted) via some external system outside the crate, and the API doesn't confuse the more regular usage.

I don't have any other oustanding comments. @fintelia ?

src/decoder/mod.rs
```diff
@@ -1389,93 +1157,111 @@ impl<R: Read + Seek> Decoder<R> {
...
-        let tile_attrs = self.image().tile_attributes.as_ref().unwrap();
-        let (padding_right, padding_down) = tile_attrs.get_padding(tile);
+        Ok(result.0)
+    }
 
-        let tile_width = tile_attrs.tile_width - padding_right;
-        let tile_length = tile_attrs.tile_length - padding_down;
+    /// Read the specified tile (at index `tile_index`) and return the binary data as a Vector and
+    /// `ChunkInfo` describing the size of the tile.
+    pub fn read_tile_at(&mut self, tile_index: usize) -> TiffResult<(DecodingResult, ChunkInfo)> {
```
fintelia, line 1204: Could we change this function to `read_chunk_at` and get rid of `read_strip_at` above?

src/decoder/mod.rs
```diff
@@ -1223,116 +1061,46 @@ impl<R: Read + Seek> Decoder<R> {
...
-                    usize::try_from(self.image().width)?,
-                    self.colortype()?,
-                )?;
-            }
-            Predictor::__NonExhaustive => unreachable!(),
-        }
         Ok(())
     }
 
     fn read_tile_to_buffer(
```
fintelia, line 1083: Could you make this just be `read_chunk_to_buffer` and have `read_strip_to_buffer` call it?

src/decoder/image.rs
```diff
@@ -77,6 +61,39 @@ impl TileAttributes {
     }
 }
 
+#[derive(Debug)]
+#[non_exhaustive]
+/// ChunkInfo describes the properties of a chunk (either tile or strip).
+pub struct ChunkInfo {
```
fintelia, line 67: This is really `pub(crate)`, right?

fintelia (changes requested):

src/decoder/image.rs
fintelia: I don't see any `pub use ChunkInfo`, and [this module isn't marked `pub`](https://github.com/image-rs/image/blob/04052e64c9a94606efc8bd3d87d5f3e0f566774e/src/lib.rs#L257).  Taken together, the struct isn't actually publicly exposed.

fintelia: I don't see any pub use ChunkInfo, and [this module isn't marked pub](https://github.com/image-rs/image-tiff/blob/master/src/decoder/mod.rs#L19). Taken together, the struct isn't actually publicly exposed.

(I think this is a good thing; if were guaranteeing a stable API for this struct, then I'd want to iterate on the interface)

src/error.rs
```diff
@@ -56,6 +56,7 @@ pub enum TiffFormatError {
         actual_samples: usize,
         required_samples: usize,
     },
+    InvalidChunkIndex(usize),
```
fintelia, line 59: This should be a `UsageError`, not a `TiffFormatError`

src/decoder/image.rs
```diff
@@ -383,13 +416,54 @@ impl Image {
         }
     }
 
+    pub(crate) fn chunk_info(&self, chunk_index: usize) -> Option<ChunkInfo> {
+        match &self.chunk_type {
+            ChunkType::Strip => {
+                let rows_per_strip = self.strip_decoder.as_ref()?.rows_per_strip as usize;
```
fintelia, line 422: The `?` should be an unwrap since if chunk_type is ChunkType::Strip then strip_decoder should always be Some.

src/decoder/image.rs
```diff
@@ -383,13 +416,54 @@ impl Image {
...
+                Some(ChunkInfo {
+                    chunk_width: sized_width,
+                    chunk_height: rows_per_strip,
+
+                    data_width: sized_width,
+                    data_height: strip_height,
+                })
+            }
+            ChunkType::Tile => {
+                let tile_attrs = self.tile_attributes.as_ref()?;
```
fintelia, line 443: `?` -> `.unwrap()`

src/decoder/mod.rs
```diff
@@ -1066,54 +1164,98 @@ impl<R: Read + Seek> Decoder<R> {
...
 
-        let strip_height = cmp::min(
-            rows_per_strip,
-            usize::try_from(self.image().height)? - index * rows_per_strip,
-        );
+    /// Read the specified strip (at index `strip_index`) and return the binary data as a Vector and
+    /// `ChunkInfo` describing the size of the strip.
+    ///
+    #[deprecated = "Use read_chunk instead"]
+    pub fn read_strip_at(&mut self, strip_index: usize) -> TiffResult<(DecodingResult, ChunkInfo)> {
```
fintelia, line 1185: Wait, you're adding this method and deprecating it at the same time?

src/decoder/mod.rs
```diff
@@ -1066,54 +1164,98 @@ impl<R: Read + Seek> Decoder<R> {
...
 
-        let byte_order = self.reader.byte_order;
-        self.image.expand_chunk(
-            &mut self.reader,
-            result.as_buffer(0),
-            tile_width,
-            byte_order,
-            tile,
-        )?;
+        self.read_tile_to_buffer(result.as_buffer(0), tile_index, chunk_info.data_width)?;
```
fintelia, line 1227: This should call `read_chunk_to_buffer`

src/decoder/mod.rs
```diff
@@ -1066,54 +1164,98 @@ impl<R: Read + Seek> Decoder<R> {
...
-        let (padding_right, padding_down) = tile_attrs.get_padding(tile);
+        Ok(result.0)
+    }
 
-        let tile_width = tile_attrs.tile_width - padding_right;
-        let tile_length = tile_attrs.tile_length - padding_down;
+    /// Read the specified tile (at index `tile_index`) and return the binary data as a Vector and
+    /// `ChunkInfo` describing the size of the tile.
+    #[deprecated = "Use read_chunk_at instead"]
+    pub fn read_tile_at(&mut self, tile_index: usize) -> TiffResult<(DecodingResult, ChunkInfo)> {
```
fintelia, line 1217: Again, this method is being added and deprecated at the same time.

src/decoder/mod.rs
```diff
@@ -1013,6 +1064,53 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(())
     }
 
+    #[deprecated = "Use read_chunk_to_buffer instead"]
```
fintelia, line 1067: No need to add a deprecation notice on a private function

src/decoder/mod.rs
```diff
@@ -1066,54 +1164,98 @@ impl<R: Read + Seek> Decoder<R> {
...
+        Ok((result, chunk_info))
+    }
+
+    /// Read a single chunk from the image and return it as a Vector. This method does not return
+    /// information on the size of the chunk read and so the caller must take care of chunks on the
+    /// rightmost column and the bottommost row in the image as the data read may be less than the
+    /// expected chunk size. This interface may change at a future date to reflect this.
+    ///
+    /// `read_chunk_at` can be used to get both the chunk data and the tile size.
+    pub fn read_chunk(&mut self) -> TiffResult<DecodingResult> {
```
fintelia, line 1238: Do we need this method? We could recommend always using `read_chunk_at` instead.

fintelia:
Added a bunch of comments, this PR is getting close! 

One other thing I do notice is that this PR uses `usize` in a bunch of places that I think it actually shouldn't be. Our public API uses `u32`'s for [image dimensions](https://docs.rs/tiff/0.7.2/tiff/decoder/struct.Decoder.html#method.dimensions) and [strip](https://docs.rs/tiff/0.7.2/tiff/decoder/struct.Decoder.html#method.strip_count) / [tile](https://docs.rs/tiff/0.7.2/tiff/decoder/struct.Decoder.html#method.tile_count) counts.

src/decoder/mod.rs
AlanRace: Ah I thought this was already in main...sorry it's been a while since I looked at the code!

fintelia:
> it seems that internally usize is used in some situations (TileAttributes for example) - so which functions should use u32 and which would you expect to be usize?

We're somewhat sloppy with internal code since it is easy to change it later. User facing code should usually used fixed size integers unless specifically dealing with something constrained by the memory on the system.

> Currently read_strip_to_buffer is a public function, but there is no tile equivalent. We now have read_chunk_to_buffer, but this is private. Should we leave this as it is? I added a deprecated message to read_strip_to_buffer, but as read_chunk_to_buffer is private, it doesn't make any sense in its current form..

I think it would make sense to only have `read_chunk` and `read_chunk_to_buffer` with both taking the chunk index as an argument. (i.e. rename `read_chunk_at` -> `read_chunk` and make the current `read_chunk_to_buffer` public). It should be fine to deprecate all the existing strip/tile functions in favor of those two.

fintelia:
How would you feel about replacing `ChunkInfo` with standalone methods on the decoder? I think that would make read_chunk_to_buffer easier to use and overall make the API a bit cleaner.
```rust
/// Returns the dimensions of a chunk.
fn chunk_dimensions(&self) -> (u32, u32) { ... }

/// Returns the dimensions `chunk`, excluding padding. Panics if `chunk` is larger than `num_chunks`.
fn chunk_data_dimensions(&self, chunk: u32) -> (u32, u32) { ... }
```

AlanRace:
I personally preferred the previous API, as everything is returned in a single function call, but I am also okay with this.

src/decoder/mod.rs
```diff
@@ -1066,61 +1139,58 @@ impl<R: Read + Seek> Decoder<R> {
...
-        )?;
+        Ok(result)
+    }
 
-        self.current_chunk += 1;
+    /// Returns the default chunk size for the current image. Any given chunk in the image is at most as large as
+    /// the value returned here. For the size of the data (chunk minus padding), use `chunk_data_dimensions`.
+    pub fn chunk_dimensions(&self) -> TiffResult<(u32, u32)> {
+        self.image().chunk_dimensions()
+    }
```
fintelia, line 1181:
```suggestion
    pub fn chunk_dimensions(&self) -> (u32, u32) {
        self.image().chunk_dimensions().unwrap()
    }
```

src/decoder/mod.rs
```diff
@@ -1066,61 +1139,58 @@ impl<R: Read + Seek> Decoder<R> {
...
 
-        Ok(result)
+    /// Returns the size of the data in the chunk with the specified index. This is the default size of the chunk,
+    /// minus any padding.
+    pub fn chunk_data_dimensions(&self, chunk_index: u32) -> TiffResult<(u32, u32)> {
+        self.image().chunk_data_dimensions(chunk_index)
     }
```
fintelia, line 1187:
```suggestion
    pub fn chunk_data_dimensions(&self, chunk_index: u32) -> (u32, u32) {
        self.image().chunk_data_dimensions(chunk_index).expect("invalid chunk_index")
    }
```

fintelia (approved):
I'd like to try to avoid returning TiffResult's from any new metadata functions if we can. With some internal refactoring we could demonstrate that chunk_dimensions is infallible and chunk_data_dimensions only fails for chunk_index>num_chunks. But for now, I'm content just adding unwrap/expect

PR #165, opened 2022-06-23
Floating point predictor

fintelia:
This change continues the work by @TomasKralCZ in #135 to add support for floating point predictors as described in http://chriscox.org/TIFFTN3d1.pdf. 

In particular, this PR:

1. Merges `fix_endianness` and `rev_hpredict` since the floating point predictor must do both at once.
2. Orders the predictor logic before `invert_colors`. I believe this doesn't matter for the horizontal predictor, but it is definitely needed for the floating point predictor.
3.  Adds the actual floating point predictor logic
4. Adds the test cases provided by @TomasKralCZ


src/decoder/image.rs
```diff
@@ -443,10 +474,33 @@ impl Image {
...
                 super::invert_colors(&mut buffer.subrange(0..total_samples), color_type);
             }
+        } else if padding.0 > 0 && self.predictor == Predictor::FloatingPoint {
+            // The floating point predictor shuffles the padding bytes into the encoded output, so
+            // this case is handled specially when needed.
+            for row in 0..(chunk_dimensions.1 - padding.1) {
+                let row_start = row * output_width * samples;
+                let row_end = row_start + (chunk_dimensions.0 - padding.0) * samples;
+
+                let mut encoded = vec![0u8; chunk_dimensions.0 * samples * byte_len];
```
HeroicKatora, line 493: Looks like this buffer can be reused across rows? Do we need to specially handle the degenerate case of no loop iterations for performance?

src/decoder/mod.rs
```diff
@@ -359,95 +359,76 @@ impl Wrapping for i64 {
...
+            match image {
+                DecodingBuffer::U8(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::U16(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::U32(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::U64(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::I8(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::I16(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::I32(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::I64(buf) => rev_hpredict_nsamp(buf, samples),
+                DecodingBuffer::F32(_) | DecodingBuffer::F64(_) => unreachable!(),
```
HeroicKatora, line 419: I'm very slightly uneasy about these assertions. These are strongly coupled with the accepted combinations in the decoder and we're one misstep in the check of decoder/image.rs:400 away from hitting this path. Can we avoid this coupling-at-adistance without too much overhead? Or may we just do a noop here, accepting a potentially wrong result in favor of stability.

fintelia: There's no especially elegant option here. Reporting the user-facing error here would require passing more context to this function that otherwise wouldn't be needed. And failing silently has its own downsides. The nice thing about panicking (especially with the added error messages) is that it is clear to a fuzzer/end users that there is a bug in the library.

However, I don't feel very strongly about this, so if there's a different option you'd prefer I'd be happy to change it.

HeroicKatora: Just wondering here, with a message this seems good enough for the moment. Maybe we'll see a way to revisit this, and as you say, will have fuzzers. Although the fuzzer coverage very much depends on which code we're fuzzing because the edge coverage is not sufficient for this: we don't invoke the code reading into a pre-allocated buffer afaik.

HeroicKatora (approved):

Issue #168, opened 2022-07-27
Panic decoding malformed tiff


micahsnyder:
The `unwrap` in `decoder/image.rs:L348` (https://github.com/image-rs/image-tiff/blob/802949f6e192073ae12c29c3c15bcabe7810f1d3/src/decoder/image.rs#L348) can fail, and crash whatever program is decoding the TIFF. Please add error handling to remove the unwrap. 

I can provide a test file, if you need to test it. I would prefer to email or message it privately. 

Regards,
Micah

micahsnyder:
Also worth mentioning, while I"m here, 0.7.3 doesn't have a git tag https://github.com/image-rs/image-tiff/tags

PR #170, opened 2022-08-02
Removed `unwrap` and set colour transform with Jpeg compressed tiffs

AlanRace:
This pull request includes better handling of jpeg compressed tiff files. The `unwrap` is removed, and instead an error (`TiffUnsupportedError::UnsupportedJpegFeature`) created and propogated back.

This also removes the hacky `add_app14segment` for setting which colour transform the `jpeg-decode` crate should expect and decode. Now the new `set_color_transform` is used based on the `PhotometricInterpretation` set in the tiff file.

Requires merging of https://github.com/image-rs/jpeg-decoder/pull/254 and https://github.com/image-rs/jpeg-decoder/pull/255 in the `jpeg-decoder` crate.

This fixes https://github.com/image-rs/image-tiff/issues/168

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -4,6 +4,8 @@ use std::io;
 use std::str;
 use std::string;
 
+use jpeg::UnsupportedFeature;
+
 use crate::decoder::{ifd::Value, ChunkType};
 use crate::tags::{
     CompressionMethod, PhotometricInterpretation, PlanarConfiguration, SampleFormat, Tag,
@@ -146,6 +148,10 @@ pub enum TiffUnsupportedError {
     UnsupportedBitsPerChannel(u8),
     UnsupportedPlanarConfig(Option<PlanarConfiguration>),
     UnsupportedDataType,
+    UnsupportedJpegFeature(UnsupportedFeature),
+    #[doc(hidden)]
+    /// Do not match against this variant. It may get removed.
+    __NonExhaustive,
 }
 
 impl fmt::Display for TiffUnsupportedError {
@@ -191,6 +197,10 @@ impl fmt::Display for TiffUnsupportedError {
                 write!(fmt, "Unsupported planar configuration “{:?}”.", config)
             }
             UnsupportedDataType => write!(fmt, "Unsupported data type."),
+            UnsupportedJpegFeature(ref unsupported_feature) => {
+                write!(fmt, "Unsupported JPEG feature {:?}", unsupported_feature)
+            }
+            __NonExhaustive => unreachable!(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


micahsnyder:
@AlanRace is this simply pending until https://github.com/image-rs/jpeg-decoder/pull/254 is merged and published in a new jpeg-decoder release?

AlanRace:
@micahsnyder yes, and the same with https://github.com/image-rs/jpeg-decoder/pull/255. It seems that https://github.com/image-rs/jpeg-decoder/pull/254 is now merged, so just waiting on https://github.com/image-rs/jpeg-decoder/pull/255 now.

AlanRace:
Now that the required changes have been merged into the latest version of jpeg-decoder, I think this is now ready for review.

src/error.rs
```diff
@@ -310,5 +324,22 @@ impl From<LzwError> for TiffError {
...
+        match error {
+            jpeg::Error::Format(format) => {
+                TiffError::FormatError(TiffFormatError::Format(format!("JPEG format: {}", format)))
+            }
+            jpeg::Error::Unsupported(unsupported) => TiffError::UnsupportedError(
+                TiffUnsupportedError::UnsupportedJpegFeature(unsupported),
+            ),
+            jpeg::Error::Io(io_error) => TiffError::IoError(io_error),
+            jpeg::Error::Internal(internal) => {
+                panic!("According to jpeg create, this is not used, so this shouldn't be reached..")
```
HeroicKatora, line 338: We should still report it, waiting on the version bump to actually remove the variant. Let's not rely on what looks like an implementation detail here. Let's report it as a `FormatError` with some opaque wrapper around the jpeg error. Actually, it would make sense to have that wrapper for both this and the `jpeg::Error::Format` variant without further discriminating between the two.

```rust
#[derive(Debug)]
struct JpegDecoderError {
    inner: jpeg::Error,
}

// and a display impl to replace eager `format!("JPEG format:  {}", error)`
// Note: Display for jpeg::Error already includes "invalid JPEG format: {}"
```

HeroicKatora (changes requested):
LGTM with regards to the mechanism. Error reporting could be improved.

src/error.rs
AlanRace: Do you mean like the version I just pushed?

If so, this would require implementing `Clone` and `PartialEq` on `jpeg::Error` to be able to be merged.

HeroicKatora: Mostly, `PartialEq` can always be added manually, `Clone` could be solved by holding the error within an `Arc` if necessary. Does that sound like an adequate solution? I'm not entirely certain.

src/error.rs
```diff
@@ -324,20 +327,32 @@ impl From<LzwError> for TiffError {
...
+#[derive(Debug, Clone, PartialEq)]
+struct JpegDecoderError {
+    inner: jpeg::Error,
+}
+
+impl From<jpeg::Error> for JpegDecoderError {
+    fn from(error: jpeg::Error) -> Self {
+        Self { inner: error }
+    }
+}
```
HeroicKatora, line 339: Please make sure to remove this `impl` and construct the error with a private converter. Otherwise, upgrading the `jpeg_decoder` dependency is a SemVer visible detail.

src/error.rs
AlanRace: I think it is probably fine, and can always be cleaned up later if desired. Latest push should include your changes and be mergable without changes to jpeg-decoder

HeroicKatora (approved):

PR #172, opened 2022-09-02
Add check to verify start index within size of buffer

### LINKED ISSUE #171: Panic decoding tiff malformed (again)
This appears to be different than https://github.com/image-rs/image-tiff/issues/168

A couple of my users have reported a crash in the image-tiff decoder:
```
thread '<unnamed>' panicked at 'range start index 16380 out of range for slice of length 9660', /Users/davidlaxer/clamav-0.105.1/libclamav_rust/.cargo/vendor/tiff/src/decoder/mod.rs:131:72
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
fatal runtime error: failed to initiate panic, error 5
```

It appears to me that `DecodingResult::as_buffer()` is blindly trusting the start offset as being within the buffer.  My guess is that `start` value is read from the file and should be verified before using it. 

If you require it for testing, please let me know.  I don't have a copy of the file to test with or share (yet).  Hopefully I will get a sample soon.

micahsnyder:
The chunk dimensions are untrusted user input and so the buffer_offset
in `read_input()` must be verified as being within the buffer or else
`result.as_buffer()` may panic.

Fixes #171

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1224,6 +1224,13 @@ impl<R: Read + Seek> Decoder<R> {
             let x = chunk % chunks_across;
             let y = chunk / chunks_across;
             let buffer_offset = y * strip_samples + x * chunk_dimensions.0 as usize * samples;
+            if buffer_offset * result.as_buffer(0).byte_len()
+                > result.as_buffer(0).as_bytes_mut().len()
+            {
+                return Err(TiffError::FormatError(
+                    TiffFormatError::InconsistentSizesEncountered,
+                ));
+            }
             let byte_order = self.reader.byte_order;
             self.image.expand_chunk(
                 &mut self.reader,


fintelia:
We validate/adjust the chunk_dimensions here:https://github.com/image-rs/image-tiff/blob/2ec1ab7489763edc9bda906c0ca7f34f2104374d/src/decoder/mod.rs#L1200-L1210

I'd like to understand why that isn't sufficient rather than just paper over the issue by returning an error if we later try to do an out of bounds array access

micahsnyder:
I assumed it was the chunk dimensions at fault because that appeared to be the value read from the file, but I wasn't sure.  I'd be happy to adjust the PR if you find the issue, or else cancel it if you find a better fix.  In the meantime, I'll see if I can solve the other issue I reported.

fintelia:
Resolved via #174 instead.

PR #174, opened 2022-09-04
Validate that images contain the proper number of strips

### LINKED ISSUE #171: Panic decoding tiff malformed (again)

fintelia:
Fixes #171

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -208,7 +208,11 @@ impl Image {
                 strip_decoder = Some(StripDecodeState { rows_per_strip });
                 tile_attributes = None;
 
-                if chunk_offsets.len() != chunk_bytes.len() {
+                if chunk_offsets.len() != chunk_bytes.len()
+                    || rows_per_strip == 0
+                    || u32::try_from(chunk_offsets.len())?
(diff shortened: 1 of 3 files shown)


HeroicKatora (approved):

micahsnyder:
Thanks @fintelia!

Do you want to cancel https://github.com/image-rs/image-tiff/pull/174 then?

PR #176, opened 2022-09-04
Return an error for zero-size images

### LINKED ISSUE #175: Zero-sized images
This crate in various places assumes that image width and image height are non-zero, but doesn't always validate that. This raises some questions: Are zero width and/or height TIFF files valid? And if so, should this crate handle them?

**Edit:** It seems that libtiff [doesn't support zero-width images](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_dirread.c#L4651), [zero row-per-strip sizes](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_dirread.c#L4665), or [zero height images](https://gitlab.com/libtiff/libtiff/-/blob/master/libtiff/tif_strip.c#L190),

fintelia:
Fixes #175

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -98,9 +98,9 @@ impl Image {
         let width = tag_reader.require_tag(Tag::ImageWidth)?.into_u32()?;
         let height = tag_reader.require_tag(Tag::ImageLength)?.into_u32()?;
         if width == 0 || height == 0 {
-            return Err(TiffError::FormatError(
-                TiffFormatError::InvalidDimensions(width, height),
-            ));
+            return Err(TiffError::FormatError(TiffFormatError::InvalidDimensions(
+                width, height,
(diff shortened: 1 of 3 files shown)


src/decoder/image.rs
```diff
@@ -97,6 +97,11 @@ impl Image {
 
         let width = tag_reader.require_tag(Tag::ImageWidth)?.into_u32()?;
         let height = tag_reader.require_tag(Tag::ImageLength)?.into_u32()?;
+        if width == 0 || height == 0 {
+            return Err(TiffError::FormatError(
+                TiffFormatError::InconsistentSizesEncountered,
```
HeroicKatora, line 102: I'd rather add a new error variant than using the wording of 'inconsistent'.

HeroicKatora (approved):

PR #186, opened 2022-11-02
Bump jpeg-decoder version pin to 0.3.0

micahsnyder:
The jpeg-decoder project has a new version with a number of improvements plus a fix for a possible panic.  I noticed just today that semantic versioning rules have us stuck on the older jpeg-decoder version.

Semantic versioning considers 0.2.4 to be incompatible with 0.3.X because of the leading zero, so this change bumps the pin to 0.3.0: https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility

> Versions are considered compatible if their left-most non-zero major/minor/patch component is the same.

Personally I would prefer to remove the version pins entirely, so we don't accidentally ship new versions of the tiff project pinned to older dependencies (again).  
But at a minimum I'd be happy to get this change in, and then if you can please publish another patch release for `image-tiff` I would be grateful. 🙏

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: ["1.48", stable, beta, nightly]
+        rust: ["1.61", stable, beta, nightly]
         command: [build, test]
     steps:
     - uses: actions/checkout@v2
@@ -20,7 +20,7 @@ jobs:
     - name: test
       run: >
         cargo test --tests --benches --no-default-features --features "$FEATURES"
-      if: ${{ matrix.rust != '1.48' }}
+      if: ${{ matrix.rust != '1.61' }}
       env:
         FEATURES: ${{ matrix.features }}
   rustfmt:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ edition = "2018"
 
 [dependencies]
 weezl = "0.1.0"
-jpeg = { package = "jpeg-decoder", version = "0.2.4", default-features = false }
+jpeg = { package = "jpeg-decoder", version = "0.3.0", default-features = false }
 flate2 = "1.0.20"
 
 [dev-dependencies]


HeroicKatora (approved):

HeroicKatora:
Seems to be good to bump to `0.3`. Please correct the MSRV used [in the CI scripts](https://github.com/image-rs/image-tiff/blob/master/.github/workflows/rust.yml#L12) (two occurences). Could you also add the new version used to `Cargo.toml` [like here](https://github.com/image-rs/jpeg-decoder/blob/master/Cargo.toml#L5-L7)?

> Personally I would prefer to remove the version pins entirely, so we don't accidentally ship new versions of the tiff project pinned to older dependencies (again).

No we won't. Stability is important for the silent majority and risking that this introduces incompatibilities is an understatement. One should be able to run `cargo build` downstream without much fear of breakage due to more recent versions. The effect of causing breakage from updates is that people no longer run any updates, which is even less desirable than getting the updates at a later date due to a staggered release of the dependency use.

On a terminology note, I'd avoid saying 'pinned' because that term refers is already used to refer to a lock file that fixes an exact (`==`) version as a resolution of the constraints at a point in time.

>  But at a minimum I'd be happy to get this change in, and then if you can please publish another patch release for image-tiff I would be grateful. pray

Sounds good, it will be `0.8` though as the msrv policy has apparently changed. Should be fine though.

micahsnyder:
@HeroicKatora sure I will add a commit to correct the MSRV.

I see that the main `image` project also sets the semver for jpeg-decoder and image-tiff in https://github.com/image-rs/image/blob/master/Cargo.toml.  I think we'll want to bump those versions as well, once the tiff 0.8 release is out.

micahsnyder:
@HeroicKatora How's this?

fintelia:
Crates.io would straight up [block us from publishing a new version](https://doc.rust-lang.org/cargo/faq.html#can-libraries-use--as-a-version-for-their-dependencies) if we tried to switch to wildcard dependencies.

fintelia (approved):

PR #187, opened 2022-11-04
Changelog for 0.8

HeroicKatora:


fintelia:
If we're going to make a major release, we should probably take the chance to remove the deprecated methods on `tiff::decoder::Decoder` and check whether there's any other breaking changes worth making

HeroicKatora:
What about the methods related to `read_chunk` that require the caller to track `current_chunk`? Should we already remove those as well or keep them deprecated for a time?

It doesn't seem to critical in that they don't block any effort for writing a decomposed abstraction. However, since we currently don't have it yet, it feels wrong to fully remove them outright.

fintelia:
I just created https://github.com/image-rs/image-tiff/pull/188. It turns out that `current_chunk` doesn't actually get used anymore, so I think that solves your concern?

fintelia (approved):
Might be worth unifying `Decoder::strip_count` and `Decoder::tile_count` while we have the chance, but otherwise looks good to me

PR #188, opened 2022-11-05
Remove deprecated methods on Decoder

fintelia:
Also removes `current_chunk` because it is no longer used elsewhere.

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -301,7 +301,6 @@ where
     reader: SmartReader<R>,
     bigtiff: bool,
     limits: Limits,
-    current_chunk: u32,
     next_ifd: Option<u64>,
     ifd_offsets: Vec<u64>,
     seen_ifds: HashSet<u64>,
@@ -586,7 +585,6 @@ impl<R: Read + Seek> Decoder<R> {
                 chunk_offsets: Vec::new(),
                 chunk_bytes: Vec::new(),
             },
-            current_chunk: 0,
         };
         decoder.next_image()?;
         Ok(decoder)
@@ -638,7 +636,6 @@ impl<R: Read + Seek> Decoder<R> {
         if let Some(ifd_offset) = self.ifd_offsets.get(ifd_index) {
             let (ifd, _next_ifd) = Self::read_ifd(&mut self.reader, self.bigtiff, *ifd_offset)?;
 
-            self.current_chunk = 0;
             self.image = Image::from_reader(&mut self.reader, ifd, &self.limits, self.bigtiff)?;
 
             Ok(())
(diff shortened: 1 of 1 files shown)


src/decoder/mod.rs
```diff
@@ -1037,33 +1026,6 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(u32::try_from(self.image().chunk_offsets.len())?)
     }
 
-    #[deprecated = "Use read_image instead"]
-    pub fn read_jpeg(&mut self) -> TiffResult<DecodingResult> {
-        self.read_image()
-    }
-
-    #[deprecated = "Use read_chunk_to_buffer instead"]
```
HeroicKatora, line 1045: Is `read_chunk` actually a good replacement yet? It has two additional arguments `chunk_index` and `output_width` to maintain.

src/decoder/mod.rs
```diff
@@ -1139,30 +1101,6 @@ impl<R: Read + Seek> Decoder<R> {
         }
     }
 
-    /// Read a single strip from the image and return it as a Vector. This method does not return
-    /// information on the size of the strip read and so the caller must take care of the final
-    /// strip in the image as the data read may be less than the expected strip size. This interface
-    /// may change at a future date to reflect this.
-    #[deprecated = "Use read_chunk instead"]
-    pub fn read_strip(&mut self) -> TiffResult<DecodingResult> {
```
HeroicKatora, line 1147: A bit simpler, here it's only the `chunk_index` to maintain.

HeroicKatora (commented):
> so I think that solves your concern?

Not really. My concern is that the replacement that the deprecations recommend don't appear to be (ergnomomic) replacements to me. I'd be _much_ gladder about the outright removal if we had a better migration story for downstream consumers.

HeroicKatora:
Not that we shouldn't do it regardless of migration, the breaking change does make sense because the current tracking of state isn't that well-designed either. In terms of maintenace, definitely better. I'd just be glad about a plan for replacing it.

src/decoder/mod.rs
fintelia: I'd argue that it is _better_ to have the `chunk_index` get passed in by the user. It is useless for the user to get a chunk of image data without knowing which one it is, so they have to be tracking that information anyway. The new approach avoids the risk of the user's value getting out of sync with the library internal version

src/decoder/mod.rs
fintelia: This version of the method requires the user to allocated the buffer themselves. Which means that they have to know the `output_width` of their buffer either way, and the extra argument just gives a bit more flexibility (though looking at `read_chunk_to_buffer` we should probably add a doc comment actually explaining what the arguments mean)

src/decoder/mod.rs
HeroicKatora: A good point. My view was from the side of processing a complete image but even that keeps the index implicitly in an iterator. Very well then, let's just see if someone comes up with a more useful and appropriate abstraction in the new interface.

HeroicKatora (approved):

HeroicKatora:
Quoting relevant information from the discussion:

> It is useless for the user to get a chunk of image data without knowing which one it is, so they have to be tracking that information anyway.

So, if anyone happens to run across this PR as the 'cause' of breakage then know that we are open to design ideas that reflect the above in a well-design interface.

PR #190, opened 2022-11-16
Added support for the YCbCr PhotometricInterpretation

kurtkuehnert:
I added support for the YCbCr PhotometricInterpretation. 
This works fine for my use case. 
I am not sure if this is all there was to do to get this working or whether I missed something.


fintelia:
How common is it for the three components to not have equal bit-depths?

It would also be good to include a test file if you have one.

HeroicKatora (commented):
Seems good, there shouldn't be too much else for the raw reading of the pixel matrix. Another PR could add the related tags that can customize the interpretation of YCbCr (in particular: `YCbCrCoefficients`) without which the numbers don't _really_ relate correctly to rgb values.

I'm just slightly wary if we would have to handle `ChromaSubsampling` etc to correctly decode the most of `YCbCr` images.  I *suppose* the tag is only used if we handled `PlanarConfiguration` as well, which we don't. But that's some other unknown for me, personally.

There's still a way to go until we could say to be Class Y compatible. This is a small & important first step.

HeroicKatora:
Regarding BitsPerSample:

> a YCbCr file
must have the following characteristics:
• SamplesPerPixel = 3. SHORT. Three components representing Y, Cb and Cr.
• BitsPerSample = 8,8,8. SHORT
> -- from Specification Revision 6.0

kurtkuehnert:
I am using this crate to convert orthophotos from this website: https://www.swisstopo.admin.ch/en/geodata/images/ortho/swissimage10.html. They are all YCbCr encoded. I am unsure how I would attribute the source correctly if I added one of them as an example.

Regarding BitsPerSample: 

Should I remove this check?
```rust
[y, cb, cr] if [y, y] == [cb, cr]=> Ok(ColorType::YCbCr(y)),
```

HeroicKatora:
> The correct copyrights statement is “Includes material © (2016-2018) Planet. All rights reserved”

Unfortunately, I don't think we can use them here, other than to reference them.

pka:
I've converted `tiled-jpeg-rgb-u8.tif` to YCbCr for testing with
`gdal_translate tiled-jpeg-rgb-u8.tif tiled-jpeg-ycbcr.tif -co COMPRESS=JPEG -co PHOTOMETRIC=YCBCR -co PROFILE=BASELINE`
[tiled-jpeg-ycbcr.tif.gz](https://github.com/image-rs/image-tiff/files/10347178/tiled-jpeg-ycbcr.tif.gz)

fintelia:
The CI failure seems to be a formatting issue. You should be able to fix it by running `cargo fmt`.

I notice that this PR is an API breaking change. I think that's justified, but it does mean we'll need to do a major version bump before releasing it.

kurtkuehnert:
@pka I have added a round trip test using this image. Thank you :)

HeroicKatora (approved):

fintelia (approved):

PR #197, opened 2022-12-16
Fix padding when image size is multiple of chunk size

pka:
For image size 500x500 and chunk size 100x100, current padding calculation returns (100, 100). With this PR (0, 0) is returned.


fintelia:
Good catch! A bit concerning that none of our tests noticed this. One note is that it might be slightly cleaner to write this as:
```rust
(self.tile_width - self.image_width % self.tile_width) % self.tile_width
```

pka:
I had failing tests in https://github.com/pka/georaster using this helpers internally.

fintelia (approved):

PR #199, opened 2023-01-12
Basic support for planar configuration

### LINKED ISSUE #198: InconsistentSizesEncountered error reading Tiff with planar config
When reading the attached file with `image::open` I get `Decoding(DecodingError { format: Exact(Tiff), underlying: Some(FormatError(InconsistentSizesEncountered)) })`.

[small_world.tif.gz](https://github.com/image-rs/image-tiff/files/10347258/small_world.tif.gz)

gdalinfo:
```
Driver: GTiff/GeoTIFF
Files: data/tiff/small_world.tif
Size is 400, 200
Coordinate System is:
GEOGCRS["WGS 84",
    DATUM["World Geodetic System 1984",
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    ID["EPSG",4326]]
Data axis to CRS axis mapping: 2,1
Origin = (-180.000000000000000,90.000000000000000)
Pixel Size = (0.900000000000000,-0.900000000000000)
Metadata:
  AREA_OR_POINT=Area
Image Structure Metadata:
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  (-180.0000000,  90.0000000) (180d 0' 0.00"W, 90d 0' 0.00"N)
Lower Left  (-180.0000000, -90.0000000) (180d 0' 0.00"W, 90d 0' 0.00"S)
Upper Right ( 180.0000000,  90.0000000) (180d 0' 0.00"E, 90d 0' 0.00"N)
Lower Right ( 180.0000000, -90.0000000) (180d 0' 0.00"E, 90d 0' 0.00"S)
Center      (   0.0000000,   0.0000000) (  0d 0' 0.01"E,  0d 0' 0.01"N)
Band 1 Block=400x20 Type=Byte, ColorInterp=Red
Band 2 Block=400x20 Type=Byte, ColorInterp=Green
Band 3 Block=400x20 Type=Byte, ColorInterp=Blue
```

AFAIK the image is in public domain, but if that is important I can convert an existing test image to a similar planar config.

pka:
Basic implementation for reading images in planar configuration.

Fixes #198.


src/decoder/image.rs
```diff
@@ -201,13 +216,22 @@ impl Image {
...
+                    }
+                };
                 strip_decoder = Some(StripDecodeState { rows_per_strip });
                 tile_attributes = None;
 
                 if chunk_offsets.len() != chunk_bytes.len()
-                    || rows_per_strip == 0
-                    || u32::try_from(chunk_offsets.len())?
-                        != height.saturating_sub(1) / rows_per_strip + 1
+                    || u32::try_from(chunk_offsets.len())? != num_chunks
```
fintelia, line 234: I think could be
```rust
let planes = match planar_config {
    PlanarConfiguration::Chunky => 1,
    PlanarConfiguration::Planar => samples as u32,
};

[...]

if chunk_offsets.len() != chunk_bytes.len()
    || rows_per_strip == 0
    || u32::try_from(chunk_offsets.len())?
        != (height.saturating_sub(1) / rows_per_strip + 1) * planes
{
    [...]
}
```

fintelia (commented):
This looks good to me. Could you also add handling for tiled planar images? I think it should just be a matter of updating the check at line 272 of src/decoder/image.rs.

pka:
> This looks good to me. Could you also add handling for tiled planar images? I think it should just be a matter of updating the check at line 272 of src/decoder/image.rs.

I don't knw how tiles planar images are organized. Are planes/bands row oriented or are they in every nth tile?

HeroicKatora:
The best I can come up with is: it doesn't specify. <https://web.archive.org/web/20210108174645/https://www.adobe.io/content/dam/udp/en/open/standards/tiff/TIFF6.pdf>

> Planar format. The components are stored in separate “component planes.” The
values in StripOffsets and StripByteCounts are then arranged as a 2-dimensional
array, with SamplesPerPixel rows and StripsPerImage columns. (All of the col-
umns for row 0 are stored first, followed by the columns of row 1, and so on.)
PhotometricInterpretation describes the type of data stored in each component
plane. For example, RGB data is stored with the Red components in one compo-
nent plane, the Green in another, and the Blue in another.

So: we can only recombine planes when PhotometricInterpretation is recognized, and only row-based encoding is mentioned.

fintelia:
The easiest strategy may be to generate a tiled planar TIFF using gdal_translate and see what the layout is

src/decoder/mod.rs
```diff
@@ -1155,7 +1162,12 @@ impl<R: Read + Seek> Decoder<R> {
         let chunks_across = ((width - 1) / chunk_dimensions.0 + 1) as usize;
         let strip_samples = width as usize * chunk_dimensions.1 as usize * samples;
 
-        for chunk in 0..self.image().chunk_offsets.len() {
+        let image_chunks = self.image().chunk_offsets.len() / self.image().strips_per_pixel();
```
HeroicKatora, line 1165: This has caused me some headache. It seems improper to return partial data with no indication of it occurring. At that point, wouldn't users simply reach for manual chunk iteration anyways? I've been considering: the library is already choosing a layout for the usual case with no control given to the caller. Couldn't we just decide and document how to handle planar images, too?

pka: Alternatively we could return no image, like now. The goal of this PR was extending the Decoder API to support planar-aware reading.

pka:
> The best I can come up with is: it doesn't specify. https://web.archive.org/web/20210108174645/https://www.adobe.io/content/dam/udp/en/open/standards/tiff/TIFF6.pdf
> 
> > Planar format. The components are stored in separate “component planes.” The
> > values in StripOffsets and StripByteCounts are then arranged as a 2-dimensional
> > array, with SamplesPerPixel rows and StripsPerImage columns. (All of the col-
> > umns for row 0 are stored first, followed by the columns of row 1, and so on.)
> > PhotometricInterpretation describes the type of data stored in each component
> > plane. For example, RGB data is stored with the Red components in one compo-
> > nent plane, the Green in another, and the Blue in another.
> 
> So: we can only recombine planes when PhotometricInterpretation is recognized, and only row-based encoding is mentioned. I can't find anything in imagemagick to indicate particular support for tile based images as well, but then again that decoding loop is somewhat obtuse C code to put it lightly.

Looks fine, seems like I was lucky.

pka:
> The easiest strategy may be to generate a tiled planar TIFF using gdal_translate and see what the layout is

We could also ask @rouault for support.

fintelia (approved):
I finally went and checked what gdal_translate does for tiled images. Turns out it outputs all the tiles for the first band, then all the tiles for the second band, and so on. This matches what the PR assumed, so I think this is set to go ahead and merge.

(Interestingly, gdal chose offsets for tiles that followed a different order: red channel of first tile, blue channel of first channel, etc. This doesn't impact decoding because TIFF has no guarantees on where in the file each chunk is in relation to the others)

PR #200, opened 2023-01-31
Fix capacity overflow in JpegReader

anfedotoff:
Hi!  
I've been fuzzing image-rs crate with [AFL++](https://github.com/AFLplusplus/AFLplusplus) using [this fuzz target](https://github.com/ispras/oss-sydr-fuzz/blob/master/projects/image-rs/fuzz-afl/fuzzers/fuzz_tiff.rs) and found a capacity overflow. I think it is better not to panic if we couldn't allocate memory. So, I tried to fix this case.
P.S. If it is interesting I could create a PR for adding AFL++ fuzz targets to image-rs.

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -458,6 +458,7 @@ impl Image {
         output_width: usize,
         byte_order: ByteOrder,
         chunk_index: u32,
+        limits: &Limits,
     ) -> TiffResult<()> {
         // Validate that the provided buffer is of the expected type.
         let color_type = self.colortype()?;
@@ -511,6 +512,9 @@ impl Image {
                 .ok_or(TiffError::FormatError(
                     TiffFormatError::InconsistentSizesEncountered,
                 ))?;
+        if *compressed_bytes > limits.intermediate_buffer_size as u64 {
+            return Err(TiffError::LimitsExceeded);
+        }
 
         let byte_len = buffer.byte_len();
         let compression_method = self.compression_method;
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1042,6 +1042,7 @@ impl<R: Read + Seek> Decoder<R> {
             output_width,
             byte_order,
             chunk_index,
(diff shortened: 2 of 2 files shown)


src/decoder/stream.rs
```diff
@@ -213,7 +213,14 @@ impl JpegReader {
...
     ) -> io::Result<JpegReader> {
         // Read jpeg image data
-        let mut segment = vec![0; length as usize];
+        let mut segment = Vec::new();
+        if segment.try_reserve_exact(length as usize).is_err() {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                format!("Unable to allocate memory of {} size for segment.", length),
+            ));
+        }
```
HeroicKatora, line 222: Thanks for identifying this. However, we don't want to check the global allocator running out of resources. On a lot of platforms this happens either not soon enough (i.e. Linux overcommit) or will still affect application stability regardless. That is, the checks of `try_reserve` are pretty much orthogonal to the DOS concerns from large images.

We'll want to check against and deduct from the configuration allocation limits instead: https://github.com/image-rs/image-tiff/blob/master/src/decoder/mod.rs#L303

You can figure out a way to make the attribute available as a reference and check against it here.

anfedotoff: I've looked more precisely at this case. And I found out, that [here](https://github.com/image-rs/image-tiff/blob/0da97709163fe08a093a00ce1e244ab301d82c02/src/decoder/image.rs#L517) we get the size that further would be used for memory allocation. Maybe it is a nice place to check Limints, but this code is located in expand_chunk method (Image struct). And Image doesn't now about Decoder's Limints.
So, to do any checks we need to change API. Honestly, I don't know if it worth it.

fintelia: The `expand_chunk` method is part of the internal API. I think it is absolutely reasonable to change the internal API to enable enforcement of limits

HeroicKatora:
Thanks!

PR #201, opened 2023-02-06
Add changelog and metadata for 0.9.0

HeroicKatora:
Let's do #199 in a separate release once we figure out tiling, no rush there I suppose.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,14 @@
+# Version 0.9.0
+
+New features:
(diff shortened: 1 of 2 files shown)


fintelia:
I think this may have to be `0.9` because of https://github.com/image-rs/image-tiff/pull/190

HeroicKatora:
Good catch, updated and added a line in the changelog for it.

PR #203, opened 2023-03-07
Don't panic when parsing metadata when custom compression is used

AlanRace:
Some microscopy file types, such as `.svs` (https://openslide.org/formats/aperio/) are essentially .tif files, with custom tags. I wanted to read in the metadata (width, height etc.), but this failed due to the file using a custom compression scheme (JPEG2000) that is not supported in the `image-tiff` library/.tif specification.

This PR simply adds in the ability to read in custom compression schemes and therefore also the metadata, without an error occuring.

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -109,8 +109,7 @@ impl Image {
         // Try to parse both the compression method and the number, format, and bits of the included samples.
         // If they are not explicitly specified, those tags are reset to their default values and not carried from previous images.
         let compression_method = match tag_reader.find_tag(Tag::Compression)? {
-            Some(val) => CompressionMethod::from_u16(val.into_u16()?)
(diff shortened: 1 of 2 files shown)


fintelia:
What happens if you try to read the file contents when an unsupported scheme is used? Do we still properly return an error?

AlanRace:
Good question - yes. When running the following code:

```rust
let svs_file = File::open(svs_path)?;
let mut decoder = Decoder::new(BufReader::new(svs_file))?;

println!("Dimensions: {:?}", decoder.dimensions());

let chunk = decoder.read_chunk(0)?;
```

I get:

```
Dimensions: Ok((39840, 28523))
Error: UnsupportedError(UnsupportedCompressionMethod(Unknown(33005)))
```

fintelia (approved):

Issue #207, opened 2023-06-28
CMYKA decoding


luksab:
I would like to propose adding CMYKA support to image-tiff.
For my use case, I only needed CMYKA decoding, but that has worked flawlessly with images exported from Adobe Photoshop, Affinity Photo 2 and Krita.
I am willing to contribute to the implementation of CMYKA support if the maintainers are interested in pursuing this feature.
Thank you for your time and consideration.

PR #208, opened 2023-07-23
add CMYKA

luksab:
As described in #207, this PR adds the CMYKA8 color type.


fintelia (approved):
Looks good to me. However, since `ColorType` isn't marked non-exhaustive, this is a breaking change and will have to wait for the next major release

PR #211, opened 2023-08-20
CCITT Group 4 / T.6 "fax machine" compression handling example

hotspoons:
The maintainer of the [fax crate](https://github.com/pdf-rs/fax) was gracious enough to help me get CCITT group 4/T.6 compression support working with the tiff crate (see [this](https://github.com/pdf-rs/fax/issues/4)) and he suggested I offer it as an example here. I'm just taking a guess for how you would want to include an example of this integration, so here it is!


sophie-h:
This code looks quite clean. Maybe this could just be added as a feature to `tiff`? That would make the example redundant.

fintelia:
Yeah, I'd be in favor of adding this as a feature to tiff. This is where we dispatch on which compression to use:
https://github.com/image-rs/image-tiff/blob/daa0e3eabf39f7837b470e94454a854fc143f6f0/src/decoder/image.rs#L361-L370

I do think it'll take a bit of effort though. The fax crate seems to sometimes [panic if it encounters malformed input](https://github.com/pdf-rs/fax/blob/f5e19658f011e771102cab424044c5c0e487c7ab/src/decoder.rs#L64) and it isn't immediately obvious whether we could handle incremental decompression (which helps both for perf and to avoid OOMs from specially crafted inputs)

197g:
Superseeded by #229

Issue #147, opened 2021-08-03
Support for arbitrary sample counts


Masterchef365:
For my use-case, I need to decode TIFFs which contain 20+ channels. Currently, image-tiff only supports specific `ColorType`s, although a previous (and slower) version of the crate did support `ColorType::Other`. I have an extremely messy patch at https://github.com/Masterchef365/image-tiff/tree/more_channels which works for my purposes, but other users of this crate might benefit from a more refined approach.

raphi-web:
I would also benefit from this! Would love to see it implemented.

PR #216, opened 2023-09-11
Addresses #214 by allowing up to 255 samples per pixel and allowing B…

tromper:
…lackIsZero and WhiteIsZero to have up to 64 bits per sample.  Also includes decode_geotiff_images unit test.


fintelia:
#215 removed the same limit this does, but it would still be great to  merge the test. Would you mind rebasing?

src/decoder/image.rs
```diff
@@ -336,7 +336,7 @@ impl Image {
                 )),
             },
             PhotometricInterpretation::BlackIsZero | PhotometricInterpretation::WhiteIsZero
-                if self.bits_per_sample.len() == 1 =>
+                if self.bits_per_sample.len() <= 64 =>
```
weiji14, line 339: It seems like this change is still needed to support multi-band GeoTIFFs after #215 is merged (the line would become `if self.samples <= 64 =>` if rebasing on top of 4a2e76858b976ae1e9838f4b20772543e3940944).

weiji14 (commented):
> #215 removed the same limit this does, but it would still be great to merge the test. Would you mind rebasing?

Been following https://github.com/image-rs/image-tiff/issues/214 and am interested in this patch to enable reading multi-band GeoTIFFs at https://github.com/weiji14/cog3pio. Is it ok if I open another PR for this?

src/decoder/image.rs
fintelia: It is a bit more complicated because other places in the code assume that `ColorType::Gray` means only a single band

weiji14: Typically for multi-band GeoTIFFs, each band would be `ColorType::Gray` (grayscale). Are you saying that we need to check the `ColorType` for each band?

fintelia: I refreshed my memory of this code. The main place that assumes that `ColorType::Gray` means the image has only a single band is [actually in the image crate](https://github.com/image-rs/image/blob/2b513ae9a6ac306e752914f562e7d408f096ba3f/src/codecs/tiff.rs#L57-L58). Other users may be making a similar assumption, particularly because `Decoder` doesn't actually expose a way to query the number of bands.

We never actually marked `ColorType` non_exhaustive so it would require a major version bump, but I think it might make sense to add a `ColorType::Multiband(bit_depth, bands)` variant or something along those lines.

weiji14: Oh ok, you mean adding `ColorType::Multiband` to the enum at https://github.com/image-rs/image-tiff/blob/895c70a1adb45deb70c2d0be1ea1b013bfeee1bf/src/lib.rs#L22-L43? I see that this seems similar to what was proposed at https://github.com/image-rs/image-tiff/issues/147 (with a draft proof of concept implemented in https://github.com/Masterchef365/image-tiff/commit/2bcc24ca5d6ee5618179b81e979b0113b5e40d63).

I'll see if I can squeeze out some time this week to implement this. It sounds like we just need to add the `ColorType::Multiband` variant, and modify the `self.samples >= 1` branch somewhere here (under the TODO): https://github.com/image-rs/image-tiff/blob/895c70a1adb45deb70c2d0be1ea1b013bfeee1bf/src/decoder/image.rs#L350-L361

weiji14: Ok, the code was a lot more straightforward than I expected. Opened a PR at https://github.com/image-rs/image-tiff/pull/224 to work on this.

Issue #217, opened 2023-10-16
Adding Bilevel support (Gray(1) ColorType)


samsieber:
I'm very interested in using this library for ingesting some document into a document processing system I work on. I'd like to be able to handle bilevel tiffs - BW tiffs with 1 bit per sample. Like this:

```
=== TIFF directory 0 ===
TIFF Directory at offset 0x101024 (1052708)
  Subfile Type: multi-page document (2 = 0x2)
  Image Width: 2550 Image Length: 3300
  Resolution: 300, 300 pixels/inch
  Position: 0, 0
  Bits/Sample: 1
  Compression Scheme: None
  Photometric Interpretation: min-is-black
  FillOrder: msb-to-lsb
  Orientation: row 0 top, col 0 lhs
  Samples/Pixel: 1
  Rows/Strip: 3280
  Planar Configuration: single image plane
  Page Number: 0-12
  ```
  
  As I understand it, you guys are open to expanding format support in this library, right? I'd be perfectly fine just being able to convert this to a Gray(8) image for manipulation in the main image library.
  
  How hard would adding that support be? I'd be happy to work on it if someone's willing to mentor me. I've poked around a bit but I'm not very familiar with the tiff spec. I got as far as trying to read a multi-page tiff like so, but part-way through it panicked because it couldn't fill the buffer when reading, and I assume that's because it was expected 1 byte per pixel instead of one bit.
  
  ```
    let contents = std::fs::read("/Users/sam/multipage-bilevel.tif").expect("Could not read file");
    let cursor = Cursor::new(&contents);
    let mut reader = tiff::decoder::Decoder::new(cursor).unwrap();
    let mut pages = 0;
    let mut read_first_page = false;
    while !read_first_page || reader.more_images() {
        println!("Got color of {:?}", reader.colortype().unwrap());
        println!("Got dimensions of {:?}", reader.dimensions().unwrap());

        if read_first_page {
            reader.next_image().unwrap();
        } else {
            read_first_page = true;
        }
        pages += 1;
        let page = reader.read_image().unwrap();
    }
  ```

fintelia:
Could you provide the exact output and if possible the failing image? We should already support `Gray(1)` images, though it is possible you're hitting a bug or the image file uses some other feature we don't support.

samsieber:
[...]

PR #218, opened 2023-10-31
Allow for intra-byte bits_per_sample.

samsieber:
Specifically, correctly read stripped tifs for any bits_per_sample sized up to the normal maximum bits_per_sample limit. Also, correctly read tiled tiffs that have bit padding in the last tile of each row. Bit padding in more than one tile per row is not supported.

Future work could include error reporting about gaps in coverage, tests, and a separate newtype for buffer indexing (vs usize for bytes)

Addresses #217; here are the caveats/support:
* 32 bits and below should work for unsigned chunky stripped images
    * I haven't tested it with planar images, but it should work
    * This supports rgb(a) tiffs as well
* Tiled tifs have the additional restriction that the only the final tile per row of tiles can have bit padding at the end. So this means that `(samples*bits_per_sample*tile_width)%byte_len` must be 0. Since tile dimensions are supposed to be multiples of 16, that means 1,2 and 4 bits_per_sample should always work. Other values may work correctly depending on the tile size and samples_per_pixel combination.
* No validation was added or altered, except noting that horizontal prediction doesn't work for any non-byte-aligned bits_per_sample size for grayscale.
* I think good future work would be to add a newtype wrapper around usize meant to index into DecodingBuffer, and then move most of the bit-padding calculation conversion logic into there.

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -557,7 +557,7 @@ impl Image {
             | (ColorType::YCbCr(n), _)
             | (ColorType::Gray(n), _)
                 if usize::from(n) == buffer.byte_len() * 8 => {}
-            (ColorType::Gray(n), DecodingBuffer::U8(_)) if n < 8 => match self.predictor {
+            (ColorType::Gray(n), DecodingBuffer::U8(_)) if n % 8 > 0 => match self.predictor {
                 Predictor::None => {}
                 Predictor::Horizontal => {
                     return Err(TiffError::UnsupportedError(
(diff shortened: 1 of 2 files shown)


src/decoder/image.rs
```diff
@@ -624,19 +624,25 @@ impl Image {
             self.jpeg_tables.as_deref().map(|a| &**a),
         )?;
 
+        // Polyfil for usize::div_ceil added in rust 1.73+
+        fn usize_div_ceil(numerator: usize, denominator: usize) -> usize {
+            (numerator + denominator - 1) / denominator
+        }
+
```
fintelia, line 631: Could you add a comment here explaining what a "buffer unit" is? Though a shorter name might be better. Maybe call them "elements"?

src/decoder/image.rs
```diff
@@ -624,19 +624,25 @@ impl Image {
...
         if output_width == data_dims.0 as usize && padding_right == 0 {
-            let total_samples = data_dims.0 as usize * data_dims.1 as usize * samples;
-            let tile = &mut buffer.as_bytes_mut()[..total_samples * byte_len];
+            let row_buffer_units: usize = usize_div_ceil(data_dims.0 as usize * samples * self.bits_per_sample as usize, byte_len * 8);
```
fintelia, line 633: I'm nervous about overflowing a `usize` here. I know the old code isn't careful here either, but it would be good to address. (Especially now that there's an additional multiplication by bits_per_sample that makes it even easier to overflow on a 32-bit system). 

Probably the right thing to do would be to do most of the arithmetic using `u64`'s and then use `try_into` when converting back.

src/decoder/mod.rs
```diff
@@ -1055,6 +1055,19 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(())
     }
 
+    /// TODO: this over-allocates when (bits_per_sample * samples) % 8 != 0.
+    /// If we leave in per-tile bit padding, then we cannot do it without knowing the number of tiles across this image is.
+    /// 
+    /// We have two options basically for precise allocation without knowing the number of tiles across:
+    ///  * Ban tiles where tile_dim % (bits_per_sample * samples) == 0
```
fintelia, line 1062: Could you add a check if this condition isn't met, and return an unsupported error in that case?

PR #221, opened 2024-01-20
Exif support

seaeagle1:
Curious whether this is (or can be turned into) something usable for more than just me.

I've added a way to access embedded Exif tags in a sort-of-standardized way. The basic idea is that allows access when decoding images for other parsing libraries and for recoding images with the Exif tag intact. Since Exif is essentially TIFF-formatted, all the basic tools were more or less in place and I just had to tweak them a bit. The things that were really missing was support for IFD sub-directories and encoding values with a dynamically derived type.

Decoder -> has a method to gather the Exif information, store the entries and feed them into an encoder for a TIFF-file without pixel data.

Encoder -> has a similar method to decode TIFF-file with-or-without pixel data and feed the Exif entries into the encoder itself.


fintelia:
Thanks for sharing this!

I've been wanting to add better EXIF handling to _image-rs_ and being able to convert the EXIF data in a TIFF file into the same form as the EXIF chunks in PNG or JPEG would certainly help with that. It'll probably be a little while before I have a chance to work on implementing that, so I'll leave this open for now

spoutn1k:
Where is this at ? I am interested in writing Exif as well and would love to help.

spoutn1k:
IFD data structures are defined twice. In `encoder`: https://github.com/image-rs/image-tiff/blob/6dc7a266d30291db1e706c8133357931f9e2a053/src/encoder/mod.rs#L161-L161

https://github.com/image-rs/image-tiff/blob/6dc7a266d30291db1e706c8133357931f9e2a053/src/encoder/mod.rs#L565-L569

And `decoder`:

https://github.com/image-rs/image-tiff/blob/6dc7a266d30291db1e706c8133357931f9e2a053/src/decoder/ifd.rs#L699

https://github.com/image-rs/image-tiff/blob/6dc7a266d30291db1e706c8133357931f9e2a053/src/decoder/ifd.rs#L334-L338

`u16` can be subsituted for `Tag` or `Type` in the above. Why not create a `ImaegFileDirectory` object that contains the data in a `cow` to keep the possibility of having references and implementing methods to manipulate the IFD on it ?

spoutn1k:
Does this require an `image-exif` crate to be used in more than tiff ?

PR #224, opened 2024-03-04
Add ColorType::Multiband enum variant to support samples>1 per pixel

### LINKED ISSUE #147: Support for arbitrary sample counts

### LINKED ISSUE #214: Very minor changes to allow multiband GeoTiff reading

weiji14:
Building on the work at #216 to support reading multi-band TIFF images (common in GeoTIFFs).

This PR extends the work done by @tromper in #216 which added the unit test for reading a 5-band GeoTIFF. I've just added the `ColorType::Multiband` enum variant as suggested at https://github.com/image-rs/image-tiff/pull/216#discussion_r1506951296), and made sure that the decoder functions can handle this multi-band type.

Refernces:
- https://www.awaresystems.be/imaging/tiff/tifftags/samplesperpixel.html

Supersedes #216, closes #214 and resolves #147.


src/decoder/image.rs
```diff
@@ -347,14 +347,19 @@ impl Image {
...
-                if self.samples == 1 =>
-            {
-                Ok(ColorType::Gray(self.bits_per_sample))
+            PhotometricInterpretation::BlackIsZero | PhotometricInterpretation::WhiteIsZero => {
+                match self.samples {
+                    1 => Ok(ColorType::Gray(self.bits_per_sample)),
+                    _ => Ok(ColorType::Multiband {
+                        bit_depth: self.bits_per_sample,
+                        num_samples: self.samples,
+                    }),
```
weiji14, line 356: Have kept the 1-band image as `ColorType::Gray` to preserve backwards compatibility as mentioned at https://github.com/image-rs/image-tiff/pull/216#discussion_r1506951296. TIFF files with bands>1 would use the new `ColorType::Multiband` enum variant here.

src/decoder/image.rs
```diff
@@ -562,18 +567,25 @@ impl Image {
...
                     ));
                 }
             },
+            (
+                ColorType::Multiband {
+                    bit_depth: _n,
+                    num_samples: _samples,
+                },
+                _,
+            ) => {}
```
weiji14, line 584: A little unsure if this should be a separate branch, or go in one of the branches above. Not familiar with what the Horizontal and FloatingPoint predictors are.

fintelia: The predictors are filters that are applied to the pixel data before compression, to help improve the compression ratio. Multiband should be included in both the match existing match arms where Gray is.

weiji14: Ok, done at e0b9ae040fe08842edc8f58182f89a31aa0c0338, but best if you can double check it. I don't have an multi-band TIFF file with a Horizontal/FloatingPoint predictor to test this out on.

PR #227, opened 2024-04-02
Add Clone to Decoder

tristan-morris:
Added clone to image-tiff, allowing the Decoder to be cloned and re-used.

Useful for Tiffs located on remote storage, without needing decoding.

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -10,12 +10,12 @@ use crate::{ColorType, TiffError, TiffFormatError, TiffResult, TiffUnsupportedEr
 use std::io::{self, Cursor, Read, Seek};
 use std::sync::Arc;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct StripDecodeState {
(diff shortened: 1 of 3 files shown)


tristan-morris:
My motivation was being able to performantly perform concurrent reads into a geotiff. This is most problematic/slow during the initial IFD and header parse as there are many seek and reads which need to be performed. This slowness is most pronounced when a geotiff is hosted on a remote s3 bucket, and opening multiple readers would be very slow. 

I'll end up closing this PR as I've since discovered std::fs::File essentially shares the same file handle and so the seek+read breaks down very quickly as the position is moved. Not sure I agree with this design decision behind a cloned Reader sharing the OS file handle, but it makes sense. So simply adding Clone isn't so simple for the 99% use-case :)

fintelia:
Hm, that does seem like too much of a gotcha to go forward with this PR. I would like to have a better API that was more conducive to parallel decoding, but API design is hard and I haven't been able to devote much time to refactoring/improvements to this crate

PR #235, opened 2024-07-22
Read Short as u16, SignedShort as i16

### LINKED ISSUE #204: Tags of type Short are read as Unsigned
When reading a tag that fits into the offset field and is of type `SHORT` (Tiff type 3) the value is returned as `Value::Unsigned` instead of `Value::Short`.

I believe the issue lies in [line 407 of ifd.rs](https://github.com/image-rs/image-tiff/blob/aefea7786f038417632f72d861c0bf32bf6299ba/src/decoder/ifd.rs#LL407C18-L407C18).

The right behavior is what can be seen when multiple values are handled ([line 475 of ifd.rs](https://github.com/image-rs/image-tiff/blob/aefea7786f038417632f72d861c0bf32bf6299ba/src/decoder/ifd.rs#L475))

weiji14:
Previously, Short values were read incorrectly as `u32` instead of `u16`, and SignedShort values were read incorrectly as `i32` instead of `i16`.

This PR does the following:
- Add `TiffFormatError::ShortExpected` enum variant
- Correctly cast `Type::SHORT` as u16 values, and support casting Short to other unsigned types via the `into_u32_vec` and `into_u64_vec` methods.
- Correctly cast `Type::SSHORT` as i16 values

Fixes #204


fintelia:
The hard part is still [this concern](https://github.com/image-rs/image-tiff/issues/204#issuecomment-1566616975):

> This makes sense to fix, though the annoying part is figuring out whether anywhere else in the code is relying on the current behavior. Or whether it is likely to break any downstream users and thus might warrant additional caution

weiji14:
> The hard part is still [this concern](https://github.com/image-rs/image-tiff/issues/204#issuecomment-1566616975):
> 
> > This makes sense to fix, though the annoying part is figuring out whether anywhere else in the code is relying on the current behavior. Or whether it is likely to break any downstream users and thus might warrant additional caution

A `tiff::decoder::ifd::Value` would still need to be cast into a primitive type using `into_u16`, `into_u32` or some other method no? My impression was that the change here would mean a user doing `Value.into_u16` would save some memory by not going through `u16` -> `u32` -> `u16` as before, but just `u16` all the way. That said, I do agree that this would warrant a minor version bump (or whatever `cargo-semver-checks` suggests :wink:).

fintelia:
Cargo semver-checks doesn't capture behavior changes. If there's downstream code that assumes (perhaps via `if let`) that it'll get a `Value::Unsigned()` for something and we switch the decoder to produce `Value::Short()` instead, the code will stop working.

weiji14:
Ok, what would you recommend then to mitigate against downstream breakages? Would a minor version bump to v0.10.0 not suffice?

tests/encode_images.rs
```diff
@@ -55,17 +55,17 @@ fn encode_decode() {
...
+                (Tag::SamplesPerPixel, ifd::Value::Short(3)),
                 (Tag::RowsPerStrip, ifd::Value::Unsigned(3334)),
                 (Tag::StripByteCounts, ifd::Value::Unsigned(30000)),
                 (Tag::XResolution, ifd::Value::Rational(1, 1)),
                 (Tag::YResolution, ifd::Value::Rational(1, 1)),
-                (Tag::ResolutionUnit, ifd::Value::Unsigned(1)),
+                (Tag::ResolutionUnit, ifd::Value::Short(1)),
                 (Tag::Artist, ifd::Value::Ascii("Image-tiff".into())),
-                (Tag::Predictor, ifd::Value::Unsigned(1)),
+                (Tag::Predictor, ifd::Value::Short(1)),
```
weiji14, line 68: > The hard part is still [this concern](https://github.com/image-rs/image-tiff/issues/204#issuecomment-1566616975):
> 
> > This makes sense to fix, though the annoying part is figuring out whether anywhere else in the code is relying on the current behavior. Or whether it is likely to break any downstream users and thus might warrant additional caution

Looks like some tests added in #255 around November 2024 (after this PR was opened) discovered a potential area of breakage. Old behaviour is that some of these tags (e.g. Compression, PhotometricInterpretation, SamplesPerPixel, ResolutionUnit, Predictor) were incorrectly decoded as Unsigned (u32), whereas after this PR, they will be decoded as Short (u16).

fintelia (approved):
I don't think there's much more we can do to figure out potential breakage, so I'll go ahead and merge this unless there's any objections

PR #236, opened 2024-07-22
Read Byte as u8, SignedByte as i8

weiji14:
Previously, Byte values were read incorrectly as `u32` instead of `u8`, and SignedShort values were read incorrectly as `i32` instead of `i8`.

This PR does the following:
- Correctly cast `Type::BYTE` as u8 values, including in `into_u16`, `into_u32`, `into_u64` methods.
- Correctly cast `Type::SBYTE` as i8 values

Similar to #235, resolves some inconsistencies raised in #204.


weiji14:
Thanks @fintelia!

PR #238, opened 2024-07-31
WIP: add fp16 support

micahcc:
Kind of a shot in the dark, but added fp16 support for my own needs.

- need an example with prediction enabled, not sure if there is a good way to generate one
- need an example with known values (this one seems to have sane values but would be good to have a predictable value)


weiji14:
Can probably close this since #257 was merged?

Issue #237, opened 2024-07-30
LZW compression expands files


spoutn1k:
I do photography on the side. I always ask for scans in TIFF format and always get scans in uncompressed TIFF formats. Because I value my disk space, I use magick to compress the TIFFs using LZW and get 15-20% of the file size back.

I have been trying to use this crate to do just that. This is the code I used:
```rs
fn compress(tif: &std::path::Path) -> Result<(), Box<dyn Error>> {
    let photo = ImageReader::open(tif)?.with_guessed_format()?.decode()?;

    let compressed = std::fs::OpenOptions::new()
        .write(true)
        .create(true)
        .open(tif.with_extension("tiff"))?;

    let mut encoder =
        tiff::encoder::TiffEncoder::<std::fs::File, tiff::encoder::TiffKindStandard>::new_generic(
            compressed,
        )?;

    encoder.write_image_with_compression::<tiff::encoder::colortype::RGB8, tiff::encoder::compression::Lzw>(
        photo.width(),
        photo.height(),
        tiff::encoder::compression::Lzw,
        photo.as_bytes(),
    )?;

    Ok(())
}
```

(This is a quick and dirty attempt and the input files come as `.tif`, so no conflicts when I use the `.tiff` extension)

Now this script produces compressed TIFF files __BIGGER__ than uncompressed originals. Is this not crazy ?
```
103701jean_003126> ll                                                                                                                     
total 181M
drwxr-xr-x   5 jb staff  160 Jul 30 21:22 .
drwx------+ 49 jb staff 1.6K Jul 30 21:22 ..
-rw-r--r--   1 jb staff  67M Jul 30 21:22 02.lzwtiff
-rw-r--r--   1 jb staff  57M Jul 26 15:47 02.tif
-rw-r--r--   1 jb staff  44M Jul 30 21:36 02.tiff
-rw-r--r--   1 jb staff  57M Jul 30 21:22 02.unctiff
```

`.tif` is the original. `.lzwtiff` was created using `image-tiff` and LZW compression, `.unctiff` was created using `image-tiff` and uncompressed output. The `.tiff` file was created using `magick -compress lzw`.

How is this possible ? Am I doing something wrong ?

feefladder:
I think this issue belongs over at weezl, since they do the LZW compression?

spoutn1k:
Nope, the compression is fine. I found out the issue is the the lack of prediction before compression. Implemented in #240.

PR #240, opened 2024-08-03
Add horizontal predictor support

spoutn1k:
While investigating #237, I came to the realization that the crate lacked predictor support at encoding. This PR attemps to merge the feature in the most efficient way.

`TiffEncoder` gains a fluent interface `with_predictor` allowing users to easily set their preferred method. Using this predictor with the LZW compression makes its average compression ratio go from 0.85:1 (yes, the compressed image is bigger) to 1.3:1.

With those changes underway, the clunkyness of the current interface was just too strong. Having the compression as a generic type on the `ImageEncoder` methods serves no purpose other than complicating the code; methods were duplicated `x` and `x_with_compression` to account for the type. Adding the predictor to this interface was out of the question: `x` `x_with_compression` `x_with_predictor` `x_with_compression_and_predictor` ? No thanks.

`TiffEncoder` also gains a `with_compression` fluent interface to set the desired file compression.

The final interface looks as such:
```rs
let mut encoder = TiffEncoder::<std::fs::File>::new(compressed)?
        .with_predictor(Predictor::Horizontal)
        .with_compression(Compression::Lzw);

    encoder.write_image::<colortype::RGB8>(
        photo.width(),
        photo.height(),
        photo.as_rgb8().expect("Wrong image format"),
    )?;
```

```diff
-    let mut encoder = TiffEncoder::<std::fs::File>::new(compressed)?;
+    let mut encoder = TiffEncoder::<std::fs::File>::new(compressed)?
+        .with_predictor(Predictor::Horizontal)
+        .with_compression(Compression::Lzw);
+
+    encoder.write_image::<colortype::RGB8>(
-    encoder.write_image_with_compression::<colortype::RGB8, Compresion::Lzw>(
         photo.width(),
         photo.height(),
-        Compression::Lzw,
         photo.as_rgb8().expect("Wrong image format"),
+    )?;
```

This means the `*_with_compression` methods are gone from the `TiffEncoder` interface, which means breaking changes.

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -1,7 +1,7 @@
 use super::ifd::{Directory, Value};
 use super::stream::{ByteOrder, DeflateReader, LZWReader, PackBitsReader};
 use super::tag_reader::TagReader;
-use super::{fp_predict_f32, fp_predict_f64, DecodingBuffer, Limits};
+use super::{predict_f32, predict_f64, Limits};
 use super::{stream::SmartReader, ChunkType};
 use crate::tags::{
     CompressionMethod, PhotometricInterpretation, PlanarConfiguration, Predictor, SampleFormat, Tag,
@@ -64,7 +64,6 @@ pub(crate) struct Image {
     pub width: u32,
     pub height: u32,
     pub bits_per_sample: u8,
-    #[allow(unused)]
     pub samples: u16,
     pub sample_format: SampleFormat,
     pub photometric_interpretation: PhotometricInterpretation,
@@ -171,8 +170,9 @@ impl Image {
             ));
         }
 
-        // This library (and libtiff) do not support mixed sample formats.
-        if bits_per_sample.iter().any(|&b| b != bits_per_sample[0]) {
+        // This library (and libtiff) do not support mixed sample formats and zero bits per sample
+        // doesn't make sense.
+        if bits_per_sample.iter().any(|&b| b != bits_per_sample[0]) || bits_per_sample[0] == 0 {
             return Err(TiffUnsupportedError::InconsistentBitsPerSample(bits_per_sample).into());
         }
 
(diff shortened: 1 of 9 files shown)


spoutn1k:
Any feedback ? If the breaking changes are not welcome I can try and work around them but I was able to build `image-rs/image` with no issues

spoutn1k:
Fixed the grammar incompatible with rust 1.61 ! Please re-run the tests I cannot do it manually :(

spoutn1k:
Any feedback ? I am using this to compress my images using rust instead of an external tool and if it is helpful to me I'm sure it could help someone else.

kornelski:
Creating a new `Vec` per row is wasteful. You could pass `&mut Vec` to the predictor and have it append pixels directly to the final buffer.

spoutn1k:
Well maybe but the method this function is used does not allow for modification of the buffer it passes:

https://github.com/image-rs/image-tiff/blob/950811888dbc23e37e97264d9aafe3937620bcad/src/encoder/mod.rs#L467-L490

So there would be copy at some point or another. I can look into changing that but I wanted to limit my changes.

kornelski:
You're calling collect() there which creates a Vec anyway. 
Don't use flat_map + collect(). Make Vec::with_capacity() and append to it

spoutn1k:
`[T::Inner]` does not have `Clone`, so if I want to create the `Vec` outside the predictor function, I will need to have this ...

I pushed a version that does not collect and writes strips as soon as they are predicted. If returning a `Vec` is still a performance concern, I will update the `ColorType` trait to do so.

kornelski:
Can you check that the tests actually work?

I've replaced the predictor code with nothing:

```rust
Predictor::Horizontal => {
                0
}
```

and `cargo test --all` still passes.

kornelski:
This is what I mean: https://github.com/spoutn1k/image-tiff/pull/1

spoutn1k:
You are right I did not implement tests. Will do that.

spoutn1k:
While implementing the tests, something was not right and the changes you made were producing corrupted images. Implementing the tests first then adding your changes.

spoutn1k:
Images are corrupted from e00fa6ba5da2da42e137d4901f51c76b5cf7d9b1. Looks like the encoder does not like writing strip by strip.

spoutn1k:
Most of the round-trip tests fail for datatypes bigger than `u8`. After some testing and manually compressing images it seems the implementation of the "deprediction" in the decoder is the culprit.

spoutn1k:
Fixes #237 and #247 !

kornelski (approved):

kornelski:
Thank you

spoutn1k:
Thank you !

Issue #232, opened 2024-05-13
FormatError(StripTileTagConflict) with minimal example trying to write an image


Ploppz:
This is more of a request for support. Trying to use this library to create a tiff file and read it:

```rust
use tiff::decoder::Decoder;
use tiff::encoder::colortype::Gray8;
use tiff::tags::{PhotometricInterpretation, Tag};

const TIFF_PATH: &str = "a.tif";

fn main() {
    // Write
    // -----

    {
        let file = std::fs::OpenOptions::new()
            .write(true)
            .create(true)
            .append(false)
            .open(TIFF_PATH)
            .unwrap();
        let mut encoder = tiff::encoder::TiffEncoder::new(file).unwrap();

        {
            let mut directory = encoder.new_directory().unwrap();
            directory.write_tag(Tag::ImageWidth, 1u32).unwrap();
            directory.write_tag(Tag::ImageLength, 1u32).unwrap();
            directory
                .write_tag(
                    Tag::PhotometricInterpretation,
                    PhotometricInterpretation::BlackIsZero as u16,
                )
                .unwrap();
            // directory.write_tag(Tag::RowsPerStrip, 1_u32).unwrap();

            directory.finish().unwrap();
        }

        // Write image
        encoder.write_image::<Gray8>(1, 1, &[0u8; 1]).unwrap();
    }

    // Read
    // ----

    let tiff_file = std::fs::File::open(TIFF_PATH).unwrap();

    // Get TIFF file "Image Description" tag:
    let _ = Decoder::new(tiff_file).unwrap();
}
```
Error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: FormatError(StripTileTagConflict)', src/main.rs:45:37
```
That's the last line trying to decode the tiff file. The Tiff file cannot be opened by other programs either.

What am I doing wrong in creation of the file? As you can see I tried RowsPerStrip to disambiguate the layout but it still gives the same error if you uncomment that line.

fintelia:
Short answer is that `write_image` internally calls `directory.finish()` so you shouldn't call it yourself. The dimensions also get set in `write_image` though I think it is harmless to set them a second time. But really we should do an overhaul of the encoding API so that stuff like that is more obvious...

Ploppz:
Deleting the `directory.finish()` line, I still get the same error.

Deleting the lines setting ImageWidth and ImageHeight, I get error `FormatError(RequiredTagNotFound(ImageWidth))`.

[...]

Issue #222, opened 2024-02-02
Private fields and interfaces make file inspection difficult


jblachly:
My use case is relatively simple: 

I need to look for images with a particular (Tag, Value) pair and remove them.

Zooming out, it would be nice if the API supported such a use case relatively directly (for example, streaming images from a `Decoder` to `Encoder` , potentially with a filter).

Still, implementing this by hand looks to be difficult to impossible, because:
1. No way to know how many IFDs are present (I understand this is a design decision to support streaming decoding)
2. `Image` is `pub(crate)`, meaning I cannot iterate a list of tags. Alternatively, a member function to provide `Directory` would be helpful. 

Am I approaching this wrongly? Tiff libraries in other languages seem to make it relatively simple to delete Images from the IFD. Thanks in advance for thoughts or advice.

fintelia:
Yeah, the crate doesn't really expose an interface for editing existing TIFF files at the moment. You can use `get_tag` and `more_images` / `next_image` to determine which images have the specific tag, but there isn't a way to request all the other tags to encode a new TIFF with everything but that one tag

jblachly:
> Yeah, the crate doesn't really expose an interface for editing existing TIFF files at the moment. You can use `get_tag` and `more_images` / `next_image` to determine which images have the specific tag, but there isn't a way to request all the other tags to encode a new TIFF with everything but that one tag

Given the way the library is structured with Decoder and Encoder, I get the idea that _modifying existing file_ is out of scope, and that seems fine to me.

My real issue is that even duplicating a file with paired Decoder/Encoder seems impossible given that the set of tags is un-knowable.

Would it make sense to make fully pub (not just `pub(crate)`) `decoder::Image`? Or, at a minimum to have a method on `Decoder` that returns `ifd: Option<Directory>`?

I can make some of these changes but would be ideal if I design it for future upstreamability

fintelia:
Exposing the IFD would probably be the way to go. The type is already public so it would just be a matter of adding the method. I've been wanting to do a more complete overhaul of the crate API for a long time, but it is kind of a lot of work and I haven't gotten around to it...

Issue #205, opened 2023-06-09
Multipage + Tile support


Jondeen:
Hi,

I'm looking into encoding `tiled` and `multipaged` TIFF images using this library, but found that the usage of `IFD directories` was defaulting to `striped` images.

I looked into the code and [made a rough implementation](https://github.com/Jondeen/image-tiff/blob/tiled_multipage/src/encoder/mod.rs) for my own internal tools.

This is far from ready for a MR, but before I spend too much time on this:
* Is there work being done on this already?
* How should I approach the current code interfaces with regards to backwards compatibility? For now I've kept all methods and constructors intact and defaulting to striped images, and added more specific names for explicitly typed versions (e.g. `fn new_image_with_type` instead of `fn new image` which defaults to striped data).
* How would you ideally implement these types. Generics? Currently I [allowed for a struct to be shared](https://github.com/Jondeen/image-tiff/blob/9eeeaee1cf4cd801396fa481a8fa08d2f5bfb2d9/src/encoder/mod.rs#L341) for both Tiles and Stripes since it doesn't add a lot of overhead on either, but of course this needs to be condensed into something more rustic. At least the majority of the code base for implementing tiled multipages is there.

My rust is still, well, a bit rusty, but I'd be happy to contribute if any of this is helpful.

fintelia:
This isn't currently being worked on, but would be a good addition. 

I'd say the first step would be to sketch out the API here. It is worth trying to add it in a backwards compatible way, but also look for whether a better design would be possible with an API break. We've done this on the decoding side where new methods were added referring to "chunks" that could either be strips or tiles depending on the image, and then the old methods deprecated and eventually removed.

Jondeen:
```mermaid
classDiagram
    TiffEncoder --|> TiffWriter : Uses
    TiffEncoder --|> TiffKind : Uses
    DirectoryEncoder --|> TiffWriter : Uses
    DirectoryEncoder --|> TiffKind : Uses
    ImageEncoder --|> DirectoryEncoder : Uses
    ImageEncoder --|> TiffKind : Uses
    DirectoryEntry --|> TiffKind : Uses
    TiffKind <|.. TiffKindStandard : Inherits
    TiffKind <|.. TiffKindBig : Inherits

```

Jondeen:
Sorry it took a while to get back to you.

As I was looking for the path of least resistance for tearing down the default setup with using stripes, I came across a few related issues as well.

### 1. Implementing Chunks instead of Stripes
[...]

Issue #85, opened 2020-08-03
Cloud compatibility


Farkal:
I open this issue to speak about the best way to have cloud compatibility.

What does it mean ? Simply that we should be able to decode the header step by step. Because the ifd can be anywhere in the file we should be able to decode ifd one by one and set the data to decode each time.

Example:
- I get the first 1024 bytes of a file
- Get the first ifd and decode it (here we could keep the current init method that decode the header and the first ifd)
- If there is another ifd and it's outside the 1024 bytes i should be able to call next_image(new_bytes) and get an error if we can't decode the next ifd (better would be to tell the size we need = 2 + the entry count at the begining of  the IFD * 12 )

If there is a way to dynamicaly add data to cursor and fake the position we should be able to already have something working but i didn't find any way to do that. I wait for your advices about this use case before making a PR 😉

fintelia:
I'm not sure if there is a crate for it, but it should be possible to make your own struct that takes a Reader and implements Read+Seek by buffering all data that is read. Something like:
```rust
struct BufferCursor<R: Read> { 
    r: R,
    data: Vec<u8>
    cursor: usize,
}
impl Read for BufferCursor {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        while cusor >= data.len() { 
             // read data into buffer
         }
        // copy data from self.data into buf
    }
}
impl Seek for BufferCursor {
    fn seek(&mut self, ...) {
         // Update cursor, but don't modify self.data or self.r
    }
}
```
If you do implement something like this, you should totally release it on crates.io though!

The bigger question is whether `image-tiff`will actually do something sane when passed a reader like this. From skimming a bit of the code it looks plausible, but I can't promise that the decoder won't seek to the end of the stream without warning/before it is actually necessary.

Farkal:
After some investigation I found that the best way was to add a new limit var:
```rust
/// Decoding limits
#[derive(Clone, Debug)]
pub struct Limits {
    /// The maximum size of any `DecodingResult` in bytes, the default is
    /// 256MiB. If the entire image is decoded at once, then this will
    /// be the maximum size of the image. If it is decoded one strip at a
    /// time, this will be the maximum size of a strip.
    pub decoding_buffer_size: usize,
    /// The maximum size of any ifd value in bytes, the default is
    /// 1MiB.
    pub ifd_value_size: usize,
    // Allow to specify a file limit to raise an error if decoding of the data is impossible (for cloud range request purpose)
[...]

PR #242, opened 2024-08-23
Exif support attempt number 2

spoutn1k:
Rebase @seaeagle1's PR and try to pull together multiple definitions of IFDs and entries. A lot has changed from the use of common structures.


feefladder:
Hey, I saw your comment on #243 and had a look around in your code, where I saw you added generics to Image. In my pr #245, I also had a deep look at the Image struct and how it works with the decoder. Also, @fintelia said [in](https://github.com/image-rs/image-tiff/issues/232#issuecomment-2109081033) [multiple]() places that the encoding API needs an overhaul. Also in #222, someone agreed with me that private fields in the Image make inspection difficult. The main point is that I think that overhaul would mean using a more config/property type of design, rather than a generics design [e.g.](https://github.com/image-rs/image-tiff/issues/205#issuecomment-1913396405) #205. I don't know what that would mean for this PR, and it should belong in its own PR, let's say we have these "problems":

- Decoder/Encoder harmonization: The API is different, the structs are different, everything is different.
- Extensibility: This library is made to build things on top of, and sort of provides API at different levels, but imho not as clearly as inspired by "[libtiff](https://libtiff.gitlab.io/libtiff/libtiff.html): provides interfaces to image data at several layers of abstraction (and cost). At the highest level image data can be read [...] without regard for the underlying data organization, colorspace, or compression scheme. Below this high-level interface the library provides scanline-, strip-, and tile-oriented interfaces that return data decompressed but otherwise untransformed. [...] At the lowest level the library provides access to the raw uncompressed strips or tiles, returning the data exactly as it appears in the file."

Now, I think it would be a good idea to have three levels of wrapping structs that hold data:
```
/// Hold all Images, possibly decoded. Also provides convenience functions
/// Just use u64, since we're not targetting super memory-constrained things
struct TIFF {
   // ifd_offsets: Option<Vec<u64>> // was seen_ifds in Decoder, but I would like a scan_ifds() function
   images: Vec<enum {Image(Image), Offset(u64)}> // indices should match, or enum 
   bigtiff: bool,
   byte_order: ByteOrder,
}

/// IFD that contains an image as opposed to a sub-ifd (which also may contain an image, but we don't necessarily check there
struct Image<'a> {
  // useful fast-access tags
  // _not_ necessarily bigtiff or endianness
  sub_ifds: Vec<IFD>
}

/// IFD that doesn't necessarily hold all tags, or even image data
struct<'a> IFD {
  sub_ifds: Vec<IFD>, // vec ?is? a pointer, so ?no? infinity otherwise Vec<Box/Arc<IFD>>
  inner: BTreeMap<Tag, Entry>
}
```  
Additionally, I think the `Decoder` could use static methods for reading IFDs, like so:
```
impl Decoder {
  /// static method for reading in an IFD, are these on Image? If they are, I think they should be exposed there
  pub fn read_ifd<R: Read + Seek>(reader: R, offset: u64) -> IFD {todo!;}
  pub fn read_ifd_async<R: AsyncRead + AsyncSeek + Unpin> (reader: R, offset: u64) -> IFD {todo!().await}
  /// scans over the linked list of ifds without reading them
  pub fn scan_images(&mut self) {
    // same as next_ifd, but doesn't read and goes on completely
    self.ifd_offsets = todo!();
  }
  /// read in all ifds of all images
  pub fn read_image_ifds<R: Read + Seek>(&self) {
    if self.ifd_offsets.is_none() {self.scan_ifds();} //should be unnecessary
    let Some(offsets) = self.ifd_offsets else { unreachable!(); }
    let self.images = offsets.map(|offset| self.read_image_ifd(offset)).collect();
  }
}
``` 

  - Document high-level vs mid/low-level api, which - I think - would 
  - Publish `Image` for debugging(#222)/wrapping purposes(#246)
    - For this, it should be clear what `Image` is. Now, there are multiple use-cases for image:
      - (partially) decoding a tiff <- the Directory<Tag, Entry> holds offsets and/or `Value`s.
      - encoding a tiff or further processing <- the Directory<Tag, DecodedEntry> holds only `Value`s
        - How to handle sub-IFDs in this case? Should we have something like:
          ```
          pub struct DecodedIfd<'a> {
            ifd: BTreeMap<Tag, DecodedEntry>;
            sub_ifds: vec<&'a DecodedIfd>
          }
          /// An Image that has retrieved all IFD info when decoding, or for encoding 
          pub struct DecodedImage<'a> {
            sub_ifds: Vec<&'a DecodedIfd> // or a more "basic" sub-ifd type
          }
          ```
          Then DecodedEntry holds a `Value::IFD(sub_ifds.find(the_ifd))` or `Value::IFD8(vec_index)`. Allows for recursion? Or Box something?
      I would say: "the [`Image`] struct holds all necessary metadata for decoding an image from an (async) decoder" <- tying `Image` and `Decoder` closer together, basically I'm trying to avoid a big maintenance
  - More control over the decoding process: Allow reading an image without directly reading all tags (#85) or read them all (#243 #242 yay tiff magic).
    - This would also allow reading just an IFD, without needing to read in all things


It's getting all a bit messy now and I can't save drafts. Also should this be its separate issue/discussion? (big refactor + API overhaul)

spoutn1k:
> Decoder/Encoder harmonization: The API is different, the structs are different, everything is different.

That's essentially what is being attempted here. The Directory type is now at the root and used in both encoder and decoder. However it expects to be tagged with the tiff format and that is wy generics are being passed down. I would sleep well knowing them gone, but this is a draft PR to see where this can be taken.

I appreciate the amount of research that shows through your comment, but I am swamped at the moment and cannot reciprocate. Please do create a discussion and we can discuss the implementation there.

feefladder:
Hey, thanks for your reply, I'll see if I can make a discussion tomorrow. I think harmonization is a bit different than exif and this PR addresses both and my PR also overhauls internal structures. So I made this comment to separate that, after gathering what is needed for exif support. As I understand, it's also not entirely clear, but at least the following:

- smooth suppport for sub-ifds. (As I could see in your PR now, it writes to the same offset it came from?)
  - reading/writing of said sub-ifds
  - Currently, you use the decoder to read in the EXIF sub-ifd. In my async implementation, asyncness function signatures bubble up all the way, so the general idea is to have generic "decoder::read_ifd()" functions that work on buffers (or memreaders for ergonomics) also on the decoding side. Then if that is useful for encoding exif, that's great. <- the synergy 🌈 

> The Directory type is now at the root and used in both encoder and decoder.

Yes, and I'm suggesting 2 "types", images and "raw" ifds. Ill see if I can polish my post more better tomorrow.

> However it expects to be tagged with the tiff format and that is wy generics are being passed down.

In the decoder, this information is a boolean flag, rather than a generic. I think essentially bigtiffness should only have to be used when reading/writing tags or headers, after which this information is redundant (if we store everything in u64 in our structures). Also ""TileByteLength"" tag in bigtiff doesn't have to be u64 per se (I think it doesn't make sense, since you'd have a very weird tiled tiff if it has tiles so big they don't fit into u32, you've done something wrong with the tiling then), which the boolean-flag-design allows for more easily.

fintelia:
This PR is far too large, with for too much all rolled up into it to meaningfully review. The proper way to add a big feature would be starting with an issue to discuss the proposed design, then if there is favorable reception for maintainers, making a series of moderately sized PRs to implement it. Though that also assumes sufficient maintainer bandwidth, which at the moment is stretched very thin.

spoutn1k:
Why is #221 not closed then ?

fintelia:
I don't fully recall the state of that other PR, but it is more than 5x smaller than this one.

TBH I haven't figured out how to handle multi-thousand line PRs. Seems to be lose-lose regardless. Essentially every time I've tried to review such a large PR, it stalls out after several rounds of feedback, either because the author gets tired of making more and more changes (understandable!) or due to design disagreements that cannot be reconciled. Ends up being a waste of everyone's time and very demoralizing.

On the other hand, other options aren't much better. Closing such a huge PR comes across as harsh given the work that went into it, but not closing is often mistaken as an encouragement to do more work. Requesting a PR to be broken into several pieces comes across as agreeing to all the feature changes. Leaving some PR feedback at the same time as one of these other options leaves very mixed messages.

spoutn1k:
I get your point. It did seem unfair and preferrential, especially as this one is a draft for exposure on an experiment and not submitted for review. I want to add that the numbers are inflated by moving code around, but that does not change the ultimate point you rightfully make. @fintelia one of the first things I did in this PR is centralize code implemented in both the encoder and decoder. Would that be a worthwhile PR to look at ?

fintelia:
Do you mean the "Created global ifd definition" change? Haven't looked in detail, but from skimming:

* **Making `Decoder` object to be generic over normal vs bigtiff:** I don't think we should do this. It makes the API more inconvenient for callers, who now have to know in advance whether they've got a normal or bigtiff file
* **Switches `Encoder` to working on typed IFD entries rather than bytes:** Not sure I understand the motivation here. Is this to make it easier to encode nested IFDs later? 

(Also reopening the PR, so we can continue in the same PR thread if we figure out how to pivot this to a sufficiently self contained change)

197g:
Yes that would be great :tada: We have an `Entry`/`Directory` structure that is supposed to be shared between different styles of directories and thus unrelated ID lists so it would be great to find out more specifically where there is the most friction for trying to use non-`tiff::tag::Tag` keys etc.

PR #244, opened 2024-09-07
README.md: advertize JPEG decode support

rouault:
sorry if I'm wrong on this, but it is my impression looking at the code that JPEG is supported


fintelia:
I think only decoding is supported

rouault:
> I think only decoding is supported

yep, https://github.com/image-rs/image-tiff/blob/master/src/encoder/compression/mod.rs#L45 confirms that. Amending my commit

PR #255, opened 2024-11-20
Add support for iterating over all tags

### LINKED ISSUE #243: Provide a way to retrieve all tags defined within a file
Currently, it is only possible to retrieve single tags via `get_tag(...)` and `find_tag(...)`. However, one might be interested in obtaining all defined tags, for instance, when modifying the TIFF and writing it to a file again. Two possible APIs come to my mind:
* `fn get_all_tags():  impl<Iterator<Item = Tag>>`, returning all tags defined in the TIFF file, or
* `fn get_all_tags():  HashMap<Tag, ifd::Value>`, returning all tags along with their value

It would be really nice to have this addition.

gschulze:
Closes #243


src/decoder/mod.rs
```diff
@@ -895,6 +898,18 @@ impl<R: Read + Seek> Decoder<R> {
         self.get_tag(tag)?.into_string()
     }
 
+    pub fn tag_iter(&mut self) -> impl Iterator<Item = TiffResult<(Tag, ifd::Value)>> + '_ {
+        match self.image().ifd.as_ref() {
+            None => Either::Left(std::iter::empty()),
+            Some(ifd) => Either::Right(TagIter::new(
```
fintelia, line 904: Rather than constructing a TagIter here, I think we could directly return:
```rust
ifd.clone().into_iter().map(...)
```

fintelia (commented):
One question I have is whether the strip/tile offsets and lengths should be included in the iteration over all tags. On one hand it is better for consistency, but they're also likely to be larger than the other tags and not especially helpful for users

src/decoder/mod.rs
```diff
@@ -895,6 +898,18 @@ impl<R: Read + Seek> Decoder<R> {
         self.get_tag(tag)?.into_string()
     }
 
+    pub fn tag_iter(&mut self) -> impl Iterator<Item = TiffResult<(Tag, ifd::Value)>> + '_ {
+        match self.image().ifd.as_ref() {
+            None => Either::Left(std::iter::empty()),
```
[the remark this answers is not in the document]
gschulze: I have no idea, honestly. I think from a developer perspective, it is more convenient to have an empty iterator instead of having to deal with an `Option`.

tests/encode_images.rs
```diff
@@ -36,6 +36,46 @@ fn encode_decode() {
         let mut decoder = Decoder::new(&mut file).unwrap();
         assert_eq!(decoder.colortype().unwrap(), ColorType::RGB(8));
         assert_eq!(decoder.dimensions().unwrap(), (100, 100));
+
+        let mut all_tags = decoder
+            .tag_iter()
+            .filter_map(Result::ok)
+            .collect::<Vec<_>>();
+        all_tags.sort_by(|(t1, _), (t2, _)| t1.to_u16().cmp(&t2.to_u16()));
```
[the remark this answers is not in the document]
gschulze: Agree.

src/decoder/mod.rs
gschulze: I tried but failed to make the borrow checker happy with this approach. Happy to change it if you can provide a working solution.

src/decoder/mod.rs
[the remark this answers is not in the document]
fintelia: Ok, I went through the existing code. It should be impossible for `ifd` to be None here, so please just `.unwrap()` it

src/decoder/mod.rs
fintelia: I was able to get it to work with:

```rust
        self.image.ifd.as_ref().unwrap().iter().map(|(tag, entry)| {
            entry
                .val(&self.limits, self.bigtiff, &mut self.reader)
                .map(|value| (*tag, value))
        })
```

(You have to directly access the `self.image` field rather than calling the `self.image()` method to make the borrow check happy. There was a reason why the two are separate, though I don't remember what it is right now)

src/decoder/mod.rs
```diff
@@ -895,6 +898,18 @@ impl<R: Read + Seek> Decoder<R> {
         self.get_tag(tag)?.into_string()
     }
 
+    pub fn tag_iter(&mut self) -> impl Iterator<Item = TiffResult<(Tag, ifd::Value)>> + '_ {
```
fintelia, line 901: Could you also add a short doc comment explaining what this method does? Something like "Returns an iterator over all tags in the current image"

src/decoder/mod.rs
gschulze: Ah, cool, thx! I incorporated your changes and pushed the branch.

gschulze:
> You might have to merge the lest changes from main locally to get the same results?

Thanks for pointing that out. I occasionally forget about that when I'm on a fork.

src/decoder/mod.rs
gschulze: Sure!

fintelia:
Thanks!

PR #256, opened 2024-12-19
Support ZSTD compression method for reading

stschake:
ZSTD is generally recognized as superior to LZW in size and speed, particularly for decompression.
We use the self-assigned tag first defined by libtiff and also used by GDAL and others.


fintelia:
Thanks for this PR, zstd support is something I've wanted for a while!

My main concern is actually around build time. On my machine, `zstd` crate takes around 10 seconds to compile and this change increases the main `image` crate's compile time by nearly 25%.

I see two possible options:
* Check whether the `ruzstd` would work as an alternative.
* Put zstd support behind a feature flag.

stschake:
Thanks for having a look! I looked into ruzstd but it seems to have a 1.5x-2x performance penalty over the reference implementation, which I felt would diminish the benefit of zstd too much. I added a feature flag now for the zstd support.

fintelia:
Thanks! If I have a chance, I might look into using ruzstd decoder by default and having the feature opt into the faster zstd bindings. Even the slower decoder may still significantly outperform LZW

PR #257, opened 2024-12-28
Add FP16 support

micahcc:
- Adds support for floating point-16 bit images
- Use the half crate since half precision floats are nightly only 
- added several tests of different features, including prediction


src/decoder/mod.rs
```diff
@@ -303,6 +318,19 @@ fn predict_f32(input: &mut [u8], output: &mut [u8], samples: usize) {
...
 
+fn predict_f16(input: &mut [u8], output: &mut [u8], samples: usize) {
+    for i in samples..input.len() {
+        input[i] = input[i].wrapping_add(input[i - samples]);
+    }
+
+    for (i, chunk) in output.chunks_mut(2).enumerate() {
+        chunk.copy_from_slice(&u16::to_ne_bytes(u16::from_be_bytes([
+            input[i],
+            input[input.len() / 4 + i],
```
fintelia, line 329: Shouldn't this be `input.len() / 2 + i`?

micahcc: good catch copy-paste error from fp32. Unfortunately I don't have an fp16 image with prediction to test with. Wonder if I can generate one somehow. 

I'm also not entirely sure the prediction for any of the floating points make sense. Should 

```
 input[i] = input[i].wrapping_add(input[i - samples]);
```
really be operating on each byte? Shouldn't it operate on the words?

fintelia (approved):
Looks good to me.

Only concern is whether we have permission to distribute the test image. Could you share the source of the image and the license it is covered under?

micahcc:
@fintelia I removed the previous image which I had pulled from the internet and switched to a pure white image generated with GIMP. I transfer all ownership of the plane white image to the image-rs project :)

micahcc:
@fintelia how do I merge this?

micahcc:
It looks like half doesn't build with 1.61, should this be feature flagged then?

fintelia:
Should be fine to bump the MSRV to 1.70. That'll require changes in Cargo.toml and .github/workflows/rust.yml

.github/workflows/rust.yml
```diff
@@ -9,16 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust: ["1.61.0", stable, beta, nightly]
+        rust: [stable, beta, nightly]
```
fintelia, line 12: Please keep checks for the MSRV by replacing `1.61.0` -> `1.70.0` rather deleting the lines entirely

micahcc: ah ok

fintelia:
Thanks!

PR #277, opened 2025-07-07
Add a dedicated Directory type

197g:
The key motivation here is the ability to represent the tags within a
directory separate from a file. This allows us to represent directories
that are not image directories, i.e. the EXIF directory, the related GPS
directory, and other sub directories. These are defined by a 'pointer'
value in another tag, an offset into the file. They can then be read
from that immediate location.

Another PR will add methods to turn the entries of a `Directory` into `ifd::Values` by using the reader / bigtiff / byte order / limits configuration from a value, which would then be enough to make Exif and GPS directories useful.

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -333,6 +333,10 @@ impl Entry {
         }
     }
 
+    pub fn count(&self) -> u64 {
+        self.count
+    }
+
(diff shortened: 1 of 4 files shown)


197g:
Is this repository intentionally setup to use squash merges? That would be different from all the rest of the `image-rs` repos.

PR #278, opened 2025-07-08
Expose methods related to reading directory

197g:
This reworks the internals of `Decoder` as a side effect, in particular groups all attributes needed to correctly interpret `Entry` as an `ifd::Value`. The main reason is that it simplifies borrow-checking however it also simplifies some interfaces where we previously passed those values as separate arguments despite needing them together.

Newly public interfaces are:
- `ifd::Value::into_ifd_pointer`, `ifd::Value::into_ifd_vec`
- `decoder::Decoder::read_directory`
- `decoder::Decoder::read_directory_tags`
- `decoder::IfdDecoder`

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -1010,13 +1010,13 @@ impl<R: Read + Seek> Decoder<R> {
     ///
     /// for ifd in sub_ifds {
     ///     let subdir = decoder.read_directory(ifd)?;
-    ///     let subfile = decoder.with_ifd(&subdir).find_tag(Tag::SubfileType)?;
-    ///     // todo: handle the subfiles, e.g. thumbnails
+    ///     let subfile = decoder.read_directory_tags(&subdir).find_tag(Tag::SubfileType)?;
+    ///     // omitted: handle the subfiles, e.g. thumbnails
     /// }
     ///
     /// # Ok::<_, tiff::TiffError>(())
     /// ```
-    pub fn with_ifd<'ifd>(&'ifd mut self, ifd: &'ifd Directory) -> IfdDecoder<'ifd> {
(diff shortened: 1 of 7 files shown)


src/decoder/mod.rs
```diff
@@ -1148,23 +965,344 @@ impl<R: Read + Seek> Decoder<R> {
...
+        } else {
+            reader.read_u16()?.into()
+        };
+
+        for _ in 0..num_tags {
+            let (tag, entry) = match Self::read_entry(reader, bigtiff)? {
+                Some(val) => val,
+                None => {
+                    continue;
+                } // Unknown data type in tag, skip
```
fintelia, line 1171: Might make more sense as part of a standalone PR, but now that we expose `tag_iter` we should probably stop ignoring unknown tags.

I'm also somewhat questioning whether it we should be converting a sorted array into a BTreeMap rather than just doing binary searches against a flat `Vec` of entries. But maybe it doesn't make much difference and wouldn't be worth the effort of changing...

fintelia (approved):

Shnatsel:
I would appreciate an addition to the docs on `pub fn read_directory` because as a non-expert in TIFF I have no idea what this means. From the title I expected it to be a filesystem directory, but the type signatures make me think that it's not.

197g:
@Shnatsel PTAL at the added documentation?

Shnatsel:
Looks good to me, thank you!

197g:
I'm going to do a pass over the Decoder interface so that it works with the `Encoder`. And then turn to `next_*` as discussed [here](https://github.com/image-rs/image-tiff/pull/279#issuecomment-3063423827).

src/decoder/mod.rs
fintelia: I was suggesting storing the tags in a sorted `Vec` rather than a `BTreeMap` since we already have them in sorted order. Though I don't know if it would actually make any difference

PR #282, opened 2025-07-13
Refactor encoding of directories

197g:
- Transform values into `Entry` eagerly instead of buffering them, the ordering constraint of TIFF only concerns the dictionary itself; does not apply to the offsets of values. This should also prepare the code structure for greater flexibility in where values are being written. Eventually, we should be able to write values *after* a directory as some decoders expect (by rewriting a whole directory as we are rewriting the next field).
- Remove the implicit assumption of all IFDs being chained without intermediate padding / values written outside the directory encoders. Instead, the link field into which to write a directories offset in `finish` is passed explicitly. The encoder is also explicitly informed about the offset of the written directory's field.


197g:
Note: I would also like to refactor the generics based design here in time. We do not differentiate the decoder via types and the encoder should be more consistent with this. Also, I'm not content with the code relying on writing files in native byte order; it should be able to re-encode values into any chosen byte order just as the decoder can do so.

fintelia:
> Also, I'm not content with the code relying on writing files in native byte order; it should be able to re-encode values into any chosen byte order just as the decoder can do so.

I'm not necessarily opposed, but is there a particular demand for this? The TIFF format is excessively flexible and while our decoder has to handle all common variations, there's no requirement that encoding also allow it.

Though I could see an argument for always encoding in little endian to provide consistency across platforms.

197g:
Demand is a good point. In terms of priority, if it is a residual result of further cleanups and refactors I'll try to ft it in. That said, I see a few more advantages:

1. It becomes easier to test all paths o the decoder by roundtrips.
2. More faith in cross-platform behavior as noted through consistency.
3. Unit test coverage isn't great in encoding. Similar to roundtrip, it should be more robust if we could verify all behavior.
4. Compression behavior changes a little unpredictable by endianess, I think? If there's demand for cross platform consistency then this is a good reason behind it.

PR #286, opened 2025-07-14
Resolve all the clippy lints and add it to CI

197g:
As the title says. Since we have bumped our rust-version high enough, we can use `div_ceil` as a welcome readability bump. The long-standing `#[non_exhaustive] struct Limits` is the other bigger change to the API surface, that is however SemVer compatibile as far as downstream is concerned.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -45,6 +45,10 @@ jobs:
     - name: Cargo check
       run: cargo check --features "${{ matrix.feature }}" --no-default-features
 
+    - name: Cargo clippy
+      if: ${{ matrix.feature != '' }}
+      run: cargo clippy --no-deps --all-features --all-targets -- -D warnings
+
   other_archs:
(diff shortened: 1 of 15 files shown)


.github/workflows/rust.yml
```diff
@@ -45,6 +45,10 @@ jobs:
     - name: Cargo check
       run: cargo check --features "${{ matrix.feature }}" --no-default-features
 
+    - name: Cargo clippy
+      if: ${{ matrix.feature != '' }}
+      run: cargo clippy --no-deps --all-features --all-targets -- -D warnings
```
fintelia, line 50: Shouldn't this use the same features as the check build? Or else run outside of the features matrix?

197g: I assume that we want to clippy all the code. And running it here is faster, until integration of caching like https://github.com/image-rs/image/pull/2528. I'd do that in a separate PR since it will very probably churn the CI a bit.

PR #289, opened 2025-07-22
Add inner() method to decoder to provide access to raw reader

### LINKED ISSUE #288: Provide access to read raw bytes in decoder
**Would you be amenable to adding a `read_bytes(length: usize)` method to the [`tiff::decoder::Decoder`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html) struct?** This would be similar to the [`read_byte()`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html#method.read_byte) method but would allow for more optimized reads by reading in a number of bytes from the underlying reader.

Use Case:
I have a use case where I want to access the raw bytes of the underlying TIFF file. In particular I want to be able to access the raw tiles/strips without decoding them (effectively [`read_chunk`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html#method.read_chunk) but without decoding). This is a part of a file conversion process where I do not want to incur the data loss from decoding and re-encoding. Currently I am doing this by maintaining two copies of the Read + Seek object (e.g. two calls to `std::fs::File::open` or reading the file into memory and using two `std::io::Cursor`s on the data). I am able to get the tile offsets and byte counts (or strip offsets and byte counts) using the `tiff::decoder::Decoder` and then use the other reader to read the actual bytes. I can't reuse the reader I pass to `tiff::decoder::Decoder` due to ownership issues.

I did try repeated calls to [`goto_offset_u64`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html#method.goto_offset_u64) and [`read_byte`](https://docs.rs/tiff/latest/tiff/decoder/struct.Decoder.html#method.read_byte), but it is prohibitively slow.

I'm happy to submit a PR if this `read_bytes(length: usize)` method sounds like something you would support!

Thank you for your work on this great project!

aotimme:
This provides an `inner()` method to the `Decoder` struct to make it easy to access the raw stream (`Read + Seek`) object provided to `Decoder`.

Closes #288

### DIFF
--- a/src/decoder/mod.rs
+++ b/src/decoder/mod.rs
@@ -660,6 +660,11 @@ impl<R: Read + Seek> Decoder<R> {
         }
     }
 
+    /// Returns a mutable reference to the stream being decoded.
+    pub fn inner(&mut self) -> &mut R {
+        self.value_reader.reader.inner()
+    }
+
     /// Reads a TIFF byte value
     #[inline]
     pub fn read_byte(&mut self) -> Result<u8, io::Error> {
--- a/tests/decode_images.rs
+++ b/tests/decode_images.rs
@@ -229,15 +229,15 @@ fn test_tiled_rgb_u8() {
 
 #[test]
 fn test_tiled_rect_rgb_u8() {
-    let file = "tiled-rect-rgb-u8.tif";
-    let expected_type = ColorType::RGB(8);
+    test_image_sum_u8("tiled-rect-rgb-u8.tif", ColorType::RGB(8), 62081032);
+}
 
-    test_image_sum_u8(file, expected_type, 62081032);
-
-    let path = PathBuf::from(TEST_IMAGE_DIR).join(file);
+#[test]
+fn test_inner_access() {
+    let path = PathBuf::from(TEST_IMAGE_DIR).join("tiled-rect-rgb-u8.tif");
     let img_file = File::open(path).expect("Cannot find test image!");
     let mut decoder = Decoder::new(img_file).expect("Cannot create decoder");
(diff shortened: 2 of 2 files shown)


tests/decode_images.rs
```diff
@@ -228,7 +229,55 @@ fn test_tiled_rgb_u8() {
 
 #[test]
 fn test_tiled_rect_rgb_u8() {
-    test_image_sum_u8("tiled-rect-rgb-u8.tif", ColorType::RGB(8), 62081032);
+    let file = "tiled-rect-rgb-u8.tif";
+    let expected_type = ColorType::RGB(8);
+
+    test_image_sum_u8(file, expected_type, 62081032);
```
aotimme, line 235: I modeled this modification off of the test `test_planar_rgb_u8` which checks the image sum like many of the other tests but also adds some additional logic tests.

src/decoder/mod.rs
```diff
@@ -660,6 +660,11 @@ impl<R: Read + Seek> Decoder<R> {
         }
     }
 
+    /// Returns a mutable reference to the stream being decoded.
+    pub fn inner(&mut self) -> &mut R {
```
aotimme, line 664: I think this is what the consensus seemed to be from #288.

tests/decode_images.rs
fintelia: Could you make a dedicated test for this rather than extending an existing test?

aotimme: I split it out and also split out the planar band chunk reading bits from the `test_planar_rgb_u8` test into a separate test.

aotimme:
Thank you @197g and @fintelia!!

PR #291, opened 2025-08-25
Populate changelog for 0.10.1

Shnatsel:
Seems like a good time to cut a release

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,20 @@
+# Version 0.10.1
+
+New features:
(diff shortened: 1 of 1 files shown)


197g (approved):
Not completely fun fact: I spilled Mate over my laptop in the process of sketching them on the weekend, so yes, that was in the pipeline.

197g:
Regarding fax4, yes, if it is mostly integration work then it should be this week. If something unexpectedly turns up then maybe a bit more.

The data survived, laptop keyboard and some other stuff needs repair :grimacing:

PR #292, opened 2025-08-26
Changelog 0.10.2

197g:
I've already yanked `0.10.1`, apparently I've accidentally moved `find_tag_unsigned_vec` to the new `IfdDecoder` without reproducing its proxy interface in `Decoder` which made that version Semver incompatible. Adding the semver checks as we have in `image` would be neat.


fintelia (approved):

197g:
Verified that `image` compiles now with the new version for now.

fintelia:
I think it can be reused directly from the `image` repository, though moving it a dedicated repository is also an option

PR #293, opened 2025-08-29
Fix type mismatch for reading Lists of Bytes

1c3t3a:
At the moment when writing Byte strings into tiff metadata sections (like for XMP metadata) and try to read it with this crate, you will find your characters as a Value::UnsignedBig, holding the char value with padding.

This seems wrong and we should instead return actual Value::Bytes.

### DIFF
--- a/src/decoder/ifd.rs
+++ b/src/decoder/ifd.rs
@@ -577,10 +577,10 @@ impl Entry {
             Type::BYTE => self.decode_offset(self.count, bo, bigtiff, limits, reader, |reader| {
                 let mut buf = [0; 1];
                 reader.inner().read_exact(&mut buf)?;
-                Ok(UnsignedBig(u64::from(buf[0])))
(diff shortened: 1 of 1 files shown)


197g:
Honestly, that is probably not the big deal in the grand scheme The type mismatch is .. awkward but `ifd::Value` is 32 bytes regardless of whether or not the variant is `Byte` or `u64`.

197g (approved):
I'm open to reworking, for the next major version, that whole `Value` struct to something more memory friendly (including also reworking the `DecodingResult`) so that we can treat the buffer as _bytes_ and actually make a list as a list of bytes instead a list of `Value` structs, encode outputs with varying bit-width, and in general have better buffer re-use with a potential output buffer allocation. 

Hopefully, this isn't [a load bearing spacebar for someone](https://xkcd.com/1172/).

PR #294, opened 2025-08-31
Make fax a default feature

Shnatsel:
Fax was commonly requested, so let's make it Just Work when tiff crate is used through `image` where it's difficult to fiddle with tiff crate features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ criterion = "0.3.1"
 crc32fast = "1.5"
 
 [features]
-default = ["deflate", "jpeg", "lzw"]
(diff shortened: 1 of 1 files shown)


197g (commented):
I'm not sure if it is ready as such, yet. I'm glad for using `fax` and it certainly does something non-trivial with a decent implementation; but it also uses a build-time `syn=1` dependency for an internal derive feature. Before having it on as a default feature I'd like to ensure we're on a cleaner dependency tree—it's harder to turn off than on and in `image` we don't use an explicit feature list.

197g (approved):

Issue #296, opened 2025-08-31
Photometric interpretation for Fax4 / 1-bit images


197g:
The test image added in this PR has `BlackIsZero` and is decoded correctly, but the [libtiff test image](https://github.com/image-rs/image/issues/1822#issuecomment-3240324116) has `WhiteIsZero` and is decoded incorrectly.

Meanwhile ImageMagick 6.9.11 decodes both correctly, as does GIMP. 

I also could not reproduce the issue with imagemagick reportedly writing broken files. Here's the output of `convert input.png -depth 1 -compress group4 imagemagick_group4.tiff`

[imagemagick_group4.tiff](https://github.com/user-attachments/files/22068286/imagemagick_group4.tiff)

It has `WhiteIsZero` interpretation and GIMP and imagemagick both open it correctly while `tiff` crate inverts the colors.

It seems the photometric interpretation needs to be read and honored after all. Either that, or we're the only ones honoring it and THAT is causing bugs.

_Originally posted by @Shnatsel in https://github.com/image-rs/image-tiff/issues/229#issuecomment-3240361581_

197g:
Thanks for the investigation @Shnatsel, I went off the comment you had linked in the rebase. Should be a matter of switching true and false bits.

Shnatsel:
Oh, that sounds likely! Just gotta add a branch for for it that does `*x = *!x;`

And probably also add an `unreachable!()` to the catch-all case there so that it panics instead of silently doing nothing when hitting an unknown color type.

197g:
Something is weird regardless. In the main branch we have `fax34::Color::Black => 1` and an image with a photometric interpretation of `min-is-black`. That image never hits the invert block, but apparently decodes exactly as it should despite literally encoding black as max in the stream interpretation?

PR #297, opened 2025-08-31
Verify WhiteIsZero invert, add Gray(1) color type

197g:
This was found through the fax4 decoding that is always bilevel.

No longer silently ignore an unknown color type when a photometric interpretation requires an inversion. DICOM requires that WhiteIsZero and BlackIsZero are only used for samples-per-pixel=1 but we do not strictly enforce that early. Previously for the result Multiband color we just ignored the tag which is also not right.

Closes: #296

@Shnatsel the image you linked is now in the test set and `image`'s convert example gives the correct colors in a test. I'm still puzzled about the code block in fax where bits seem flipped, to me at least. Maybe fax flipped their Black/White colors in the returned stream, too? It's puzzling.


Shnatsel:
Well, it fixes decoding of imagemagick output (which you added to the repo) and the libtiff fax4 test file. And the `BlackIsZero` image committed originally still works. So good enough for me.

Shnatsel:
There is a non-fax-encoded 1-bit TIFF in the `image` test suite, and with this change it still decodes fine: https://github.com/image-rs/image/blob/main/tests/images/tiff/testsuite/l1.tiff

Shnatsel (approved):

197g:
I've verified with libtiff. It seems the old 'convention' or default behavior of treating whiteiszero might have eaten its tail in here somehwere. Here's the runlength decoding loop for libtiff: https://gitlab.com/libtiff/libtiff/-/blob/85f2ac8e0b01cb7db2bbecf4a3b891bdbef67938/libtiff/tif_fax3.c

The critical portion is that its loop always does white-black at the same time, fax starts with white. And in that loop we have:

```c
// white section
                        for (; n && !isAligned(cp, int64_t); n--)
                            *cp++ = 0x00;
[…]
// black section
                        for (; n && !isAligned(cp, int64_t); n--)
                            *cp++ = 0xff;
```

Whatever history is behind here it is consistent with treating the run that is, per ITU, white as `0` and black as `1`. I'll add comments and we can merge this.

src/decoder/mod.rs
```diff
@@ -475,11 +475,21 @@ fn fix_endianness_and_predict(
...
 
-fn invert_colors(buf: &mut [u8], color_type: ColorType, sample_format: SampleFormat) {
+fn invert_colors(
+    buf: &mut [u8],
+    color_type: ColorType,
+    sample_format: SampleFormat,
+) -> TiffResult<()> {
     match (color_type, sample_format) {
-        (ColorType::Gray(8), SampleFormat::Uint) => {
+        (ColorType::Gray(n), SampleFormat::Uint) if 8 % n == 0 => {
```
fintelia, line 484: This condition might be wrong. It doesn't cover `n=1`, but does overlap with the n=16, 32, and 64 cases below...

Though actually I think the `for x in buf { *x = !*x; }` loop should work for any `SampleFormat::Uint` value

197g: Quite the other way around, it covers n=1,2,4,8

Shnatsel:
```suggestion
        (ColorType::Gray(n), SampleFormat::Uint) if 8 % n == 0 => {
        (ColorType::Gray(1 | 2 | 4 | 8), SampleFormat::Uint) => {
```

This spells out the values explicitly

197g: For some reason I thought or-patterns beyond the top-level were quite recent. But it's 1.56. Is this grey beard status? But don't want any beard, let alone grey :crying_cat_face:

PR #330, opened 2026-01-14
Helper method to copy EXIF data

197g:
The main complexity here is that firstly the entries are a tree and secondly some entries are not-allowed or ill-advised when changing the photometric interpretation / planar configuration of the image data. So we have two use-cases here: 

- copying EXIF data while modifying the image sequence and re-writing the image.
- copying EXIF data into an independent file.

This is solved by classifying tags through a `TagFilter` type. (As an experiment, that could include a 'Personally Identifiable Data' qualification but that is not super well-defined and I'd like to externalize the legal aspect in that through future extension methods, not seeing much adoption for this use without that, too). So there's `All` for copying tags just through their type and count and `Known` to ensure we at least semantically recognize the tags as part of EXIF v3 which enables us to remove tags that would conflict with a changed photometric interpretation / planar configuration.

### DIFF
--- a/src/encoder/mod.rs
+++ b/src/encoder/mod.rs
@@ -2,27 +2,27 @@ pub use tiff_value::*;
 
 use std::{
     cmp,
-    io::{self, Seek, Write},
+    io::{self, Read, Seek, Write},
     marker::PhantomData,
     mem,
     num::{NonZeroU64, TryFromIntError},
 };
 
 use crate::{
-    decoder::{ifd::Entry, IfdDecoder},
+    decoder::{ifd::Entry, Decoder},
     error::{TiffResult, UsageError},
     tags::{
         ByteOrder, CompressionMethod, ExtraSamples, IfdPointer, PhotometricInterpretation,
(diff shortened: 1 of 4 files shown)


197g:
Todo: Exif has an unrelated `UTF-8` value type which is not standard (129).

> An 8-bit byte representing a string according to UTF-8[22]. The final byte is
terminated with NULL[00.H]. BOM (Byte Order Mark) shall not be used. The
UTF-8 count shall include NULL.

It's not clear if we should just support this for all files. Technically only allowed in EXIF private directories but I have yet to encounter this otherwise. And we could fall back to ignoring the tag when UTF-8 validation fails. I don't know yet, so it's not part of this PR. Those tags are thrown away for now.

nicopap:
I'd be interested in helping out with this PR. What relevant action can I do to help out?

197g:
It would be good to validate that the choices this allows are useful (by an actual use-case for it). There's significant complexity in it, that'll also complicate any later fixing should some of the choices be incompatible with some use. One the upside we can probably just provide a different method, as mentioned somewhere in the rationale, but a little bit of caution would still be better :)

PR #337, opened 2026-02-03
The whole libtiffpic test suite and SgiLuv compression

197g:
Unfortunately the test suite is quite big, though 5MB is still remarkably small for a lot of regression and coverage tests. I think we might cut out some of the larger files from the repository itself.

[SgiLuv](https://github.com/libsdl-org/libtiff/blob/885b36e137c584aaf784fc63784340a2c98a41d5/libtiff/tif_luv.c#L1032) is added as the test files cover it. It's straightforward to implement except that it couples the compression tag with the sample encoding directly. The raw sample values don't really mean anything. We could return CIE XYZ or CIE xyY data instead but libtiff converts to Rgb as well and we must return floating point data to avoid loss of precision. I'm not aware of any tags we could synthesize to proper defaults to describe the quantized form otherwise. We could return _raw_ raw data (i.e the compressed form) but that should be covered by another feature request.

Draft until the libtiff coverage / space tradeoff is resolved and because the hacky way of controlling the color needs a design review that allows us to be certain the code after it uses it correctly (e.g. there's probably an endian problem right now).

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -4,8 +4,8 @@ use super::tag_reader::TagReader;
 use super::ChunkType;
 use super::{predict_f16, predict_f32, predict_f64, ValueReader};
 use crate::tags::{
-    CompressionMethod, ExtraSamples, PhotometricInterpretation, PlanarConfiguration, Predictor,
-    SampleFormat, Tag,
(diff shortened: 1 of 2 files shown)


197g:
The test suite does detect a genuine error, there seems to be a problem with some forms of tiled images. The `ImageLength` tag and `ImageWidth` tags are not used properly?

fintelia (approved):

PR #388, opened 2026-04-17
feat: YCbCr chroma subsampling for non-JPEG compression

aarshivv:
Follow-up to #387. Brings back 83409cb (reverted in 85215e8) with two extra patches to address gaps.

First commit is a clean cherry-pick of 83409cb — `YCbCrUpsamplingReader` that converts MCU-block layout into full-res YCbCr triples, wired into expand_chunk for non-JPEG YCbCr. Libtiffpic hashes updated for `tests/libtiffpic/dscf0013.tif` (uncompressed 4:2:2) and `tests/libtiffpic/ycbcr-cat.tif` (LZW 4:2:0).

Second commit does two things:

- Rejects predictor != None with subsampling. The pipeline currently reverses the predictor after upsampling, but the predictor was encoded against the stored block layout — reversing it on upsampled data is wrong. Latent bug in the original that isn't exercised because both libtiffpic test files use predictor=1. Erroring out keeps the scope tight; reordering prediction to run before upsampling can be a follow-up.
- Adds unit tests for YCbCrUpsamplingReader in isolation: (2,2), (2,1), (1,2), (4,4), plus non-divisible width/height where the final block column/row gets clipped.

Planar YCbCr + subsampling stays rejected as in the cherry-pick.

cargo test, cargo clippy --all-targets, cargo test --test decode_libtiffpic all green (61/61).

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -938,15 +938,10 @@ impl Image {
 
         // Only this color type interprets the tag, which is defined with a default of (2, 2)
         if matches!(color, ColorType::YCbCr(_)) && self.chroma_subsampling != (1, 1) {
-            // The JPEG library does upsampling for us and defines its buffers correctly
-            // (presumably). All other compression schemes are not supported..
-            //
-            // NOTE: as explained in <fa225e820b96bef35f01bf4685654beeb4a8df0c> we may be better
-            // off supporting this tag by consistently upsampling, not by adjusting the buffer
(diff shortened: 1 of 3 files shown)


aarshivv:
CI failure here is unrelated — src/decoder/logluv.rs:95 trips a new clippy lint (while_let_loop, enabled by default in Rust 1.95). Main fails the same check on my machine. Happy to open a separate PR for the one-line fix.

Issue #265, opened 2025-03-02
How to write a multiband tiff?


swdmike:
I noticed there is a sample for reading multiband TIFF file in the tests. 

I am wondering if the image-tiff library currently supports encoding multiband TIFF files, or if there is a plan to support this feature in the future?

Thanks.

PR #390, opened 2026-05-09
fix: correct interpretation of extra samples with unspecified type of alpha

soyosaki:
The commit should correct how TIFF images with samples count larger than photometric samples count are interpreted. For example, if a TIFF image claims a RGB photometric interpretation, a `8 8 8 8` bits-per-sample array, and an unspecified extra samples field (i.e. `ExtraSamples = 0`), the current commit should interpret the image as an RGBA image with the extra sample be treated as the alpha channel of the image, whereas currently the crate would interpret it as a RGB image and produce a corrupted pixel map. 

The test `extra_bits_rgb` is fixed together, and it should now align with the logic of `extra_bits_gray` and `write_extra_bits_image`.

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -1190,7 +1190,7 @@ impl Image {
                 super::fix_endianness_and_predict(
                     out_row,
                     color_type.bit_depth(),
-                    samples,
+                    data_samples,
                     ByteOrder::native(),
                     predictor,
                     scratch,
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -518,7 +518,7 @@ impl Image {
         let is_alpha_extra_samples = matches!(
             self.extra_samples.as_slice(),
             [ExtraSamples::AssociatedAlpha, ..] | [ExtraSamples::UnassociatedAlpha, ..]
-        );
+        ) || self.samples > self.photometric_samples;
 
         match self.photometric_interpretation {
             PhotometricInterpretation::RGB => match self.photometric_samples {
--- a/tests/decode_images.rs
+++ b/tests/decode_images.rs
@@ -975,7 +975,7 @@ fn extra_bits_gray() {
 
 #[test]
 fn extra_bits_rgb() {
-    test_image_sum_u8("extra_bits_rgb_8b.tiff", ColorType::RGB(8), 0);
+    test_image_sum_u8("extra_bits_rgb_8b.tiff", ColorType::RGBA(8), 64);
 }
 
 #[test]


soyosaki:
The latest commit should address the decode issue related to TIFF data in RGB with more than 1 extra samples. Clippy warning appear unrelated to the changes on the current branch.

soyosaki:
Merged latest main, now all checks are passed.

197g:
Hm, I think I know what motivates this? However, the approach would make all the other checks completely redundant and I don't really want such large degrees of permissiveness to be an implicit default. The extra samples tag must only occur when it makes the sample count and we here now ignore its contents in all cases where it is potentially used at all. I don't like silently treating non-alpha data as alpha in what would likely be used as premultiplied image data.

There should be a method for retrieving the auxiliary data data is not part of the color information but this isn't it. Feel free to reopen with another approach, I do want that data to be retrievable. There have been a few other outstanding request to peak at that raw data (#265, #98). So with a more explicit design handling this more uniformly surely it would be okay to add.

soyosaki:
@197g This behavior is actually intentional to keep this lib in align with libtiff on alpha info interpretation. You may chack https://github.com/libsdl-org/libtiff/blob/732665c2c8785cec3e1f46ba9908575f0f3a8059/libtiff/tif_getimage.c#L366 , where it defaults to associated alpha when extra samples contains unspecified alpha as the first extra sample. The upcoming `DEFAULT_EXTRASAMPLE_AS_ALPHA` is by default enabled per https://github.com/libsdl-org/libtiff/blob/732665c2c8785cec3e1f46ba9908575f0f3a8059/libtiff/tiffconf.h.in#L121 , and it makes libtiff process images with 4 channels and rgb photometric interpretation but without extrasamples information as images as rgba image with associated alpha.

As stated at both the referred site of libtiff, there are several softwares (which includes at least many versions of Adobe softwares, as reported in my issue) that produces tiff images with incomplete extra samples info. In my use case, unfortunately, a large proportion of TIFF images are produced by these softwares.

If this behavior is not preferred by default for GeoTIFF users for example, probably we should gate these checks behind a default feature. The reason why I suggest that it be enabled by default is, the library's main goal appears to be parsing TIFF as regular images that upstream libraries like image-rs/image could use. And for achieving this goal, we should strive to maintain compatibility to common non-standard-conforming patterns in order to make it reliable for the image parsing task.

A slight mitigation is, we can split the conditional into two parts like libtiff, one handling the case where the first extra sample is unassociated, one handling the case where the image has rgb photometric interpretation and exactly 4 channels.

197g:
Gating it behind a configuration on the decoder seems fine (e.g. `set_default_alpha_interpretation`). Frankly a compilation flag seems insane on libtiff's part. There's no way that a shared library will ever ship in the configuration intended by all users of the software.. Well a good runtime flag on the Decoder is another matter, that is fine for compatibility. There's already 'Lenient' modes in other decoders. As long as it is not enabled by default in this library. On the other hand it sounds like the `image` integration will want to enable it.

soyosaki:
@197g Agreed. Things are oftentimes insane when it comes to these media files, especially when coming to the messy things developed by Apple and Adobe, and there's few we can do cause people still use them. A `lenient` feature, I believe, should be the most appropriate to implement. I suspect that there must have some undocumented traits that old softwares like libtiff support but we don't, so these future changes could be gated also under the feature. After the changes is merged, I will get a PR started for the `image` crate to enable this feature.

PR #391, opened 2026-05-11
Don't fail if EndOfInformation is omitted

jasinb:
Some real-world LZW-compressed TIFFs omit the trailing `EndOfInformation` code (libtiff accepts this with a warning). Don't fail with `UnexpectedEof` in this case.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -182,7 +182,7 @@ impl<R: Read> Read for LZWReader<R> {
                         return Ok(result.consumed_out);
                     }
                     return Err(io::Error::new(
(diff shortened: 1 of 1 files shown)


jasinb:
I was unsure if the correct error message should still be `UnexpectedEof` if we get `NoProgress` even with some input remaining, but I opted to keep the diff minimal. Another choice would be `InvalidData` or something.

jasinb:
The check failures seem unrelated.

197g:
This was likely caused by an issue in the lzw decoder that should have been fixed by #392. Do you have any test images to verify? At least I'm assuming that we do want to continue detecting streams that do not contain all necessary samples.

PR #394, opened 2026-05-28
Add PhotometricInterpretation unknown tag support

powturns:
Provides a mechanism to support private or extension photometric interpretations (such as CFA)

### DIFF
--- a/src/decoder/image.rs
+++ b/src/decoder/image.rs
@@ -660,12 +660,14 @@ impl Image {
                 ),
             )),
             PhotometricInterpretation::RGBPalette => Ok(ColorType::Palette(self.bits_per_sample)),
-            PhotometricInterpretation::TransparencyMask => Err(TiffError::UnsupportedError(
-                TiffUnsupportedError::InterpretationWithBits(
-                    self.photometric_interpretation,
-                    vec![self.bits_per_sample; self.samples as usize],
-                ),
-            )),
+            PhotometricInterpretation::TransparencyMask | PhotometricInterpretation::Unknown(_) => {
+                Err(TiffError::UnsupportedError(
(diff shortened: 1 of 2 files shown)


197g:
Can you share a little on how you expect this to be used or the motivating use case?

In the decoding portion of the code there is no use for the semantic type of this tag with that variant. Since there is by definition no information about the expected number of samples the code will almost always error here. It's more relevant to encoding but again we expect to be able to validate the sample count that we're being handed.

All of this points to a larger issue  of interacting with the file on a lower level with less validation that was started to be addressed with #288 and #352 but it's not clear what piece of needs for that use is being addressed with this change.

powturns:
This helps support writing out a tiff using a custom color type. My specific use case is writing out DNGs that have not been demosaiced (PhotometricInterpretation = CFA)

197g:
Would it make more sense to make this a specific encoder feature, or adjust the [ColorType](https://docs.rs/tiff/latest/tiff/encoder/colortype/trait.ColorType.html) trait we use (which I'm assuming you're filling with a custom impl?). It currently does not provide any path for a runtime input of the tags and values despite not relying in it for much more than matching sample format to the associated `Inner`. It's inflexible enough that there's a _single_ horizontal prediction method associated with the trait..

The whole trait could use an overhaul; or an _alternative_ entry point to encoding.

197g (approved):
Alas, regardless of encoding we want to allow the decoder to readout samples byte-for-byte without prediction and post-processing; the planar layouting is universal and depends only on sample depths (prediction should only depend on `SampleFormat` but I'm not entirely certain about compression tags and those in turn are entangled with PhotometricInterpretation for YCbCr for instance).

So in case we'll want to have an `decoder::Image` representation for this.

PR #396, opened 2026-06-16
Fix subtract-with-overflow panic in extended-depth gray inversion

lilith:
## Problem

`invert_colors` inverts `WhiteIsZero` grayscale by computing `max - v` per sample. For the extended bit depths promoted to `u16`/`u32` — `Gray(9..=15)` and `Gray(17..=31)` — `max` is `(1 << n) - 1`, which is *smaller* than the storage type's range. A TIFF whose stored sample carries bits above the declared bit depth makes `v > max`, so `max - v` underflows:

- **debug builds:** `panic: attempt to subtract with overflow` (`src/decoder/mod.rs:637`, the `u32` arm)
- **release builds:** a wrapped (incorrect) inverted pixel

The standard arms (`Gray(1 | 2 | 4 | 8 | 16 | 32 | 64)`) aren't affected — there `max` equals the type's full range, so the subtraction can't underflow. A fuzzer found the extended-depth case.

## Fix

Clamp out-of-range samples with `saturating_sub`:

```rust
*chunk = max.saturating_sub(v).to_ne_bytes();
```

A sample claiming to be brighter than the declared max inverts to 0; the result is always a valid in-range sample, and the well-formed path is unchanged.

### Why `saturating_sub` rather than masking

`saturating_sub` is the simplest correct option and costs nothing on the well-formed path:

- **u16 arm** lowers to `psubusw`, an SSE2-native unsigned saturating subtract (present on every x86-64 CPU). The loop autovectorizes to that single instruction — same cost as the wrapping `psubw` the current code emits; on aarch64 it's `uqsub`.
- **u32 arm:** there is no native 32-bit unsigned saturating subtract before AVX-512, so it lowers to a sign-flip + `psubd` emulation — still vectorized, and no larger than a mask-then-subtract alternative.

`invert_colors` is a streaming load-modify-store over the whole buffer, so it's memory-bandwidth bound and the ALU choice is lost in the noise either way. (Checked at baseline `-C target-cpu=x86-64`, i.e. SSE2.)

## Tests

Two fuzzer-minimized regression seeds added under `tests/fuzz_images/oor_panic/`, exercised by the existing `oor_panic` test. Without the fix they panic (`attempt to subtract with overflow`); with it they decode cleanly.


197g:
I think the code was operating under the assumption that the bit-depth would be adhered to by the decompression process before (which is sensible but definitely not the case). It does make me wonder though if the library should also clamp the channel values in the non-inverted color case. Either way this seems fine in general.

lilith:
Superseded by #399, which fixes the root cause more broadly: it clamps the promoted extended-depth sample to its declared bit depth right after the predictor, so the non-inverted color path @197g raised is covered too — and `invert_colors` needs no change. The two regression seeds moved there unchanged. Closing in favor of #399.

PR #397, opened 2026-06-22
Fix: add a lenient mode for image decoder

soyosaki:
Add an optional lenient mode that allows libtiff-compatible interpretation of extra samples with unspecified alpha. The extra samples interpretation behavior of libtiff is available at https://github.com/libsdl-org/libtiff/blob/732665c2c8785cec3e1f46ba9908575f0f3a8059/libtiff/tif_getimage.c#L366, and discussed in PR #390 .


soyosaki:
@197g This implements the lenient mode discussed in #390 and is now ready for review.

src/decoder/image.rs
```diff
@@ -155,9 +156,10 @@ pub(crate) struct ReadoutLayout {
 }
 
 impl Image {
-    pub fn from_reader<R: Read + Seek>(
+    pub fn from_reader_with_options<R: Read + Seek>(
         decoder: &mut ValueReader<R>,
         ifd: &Directory,
+        lenient: bool,
```
197g, line 162: The parameter is required and there's no longer a `from_reader`, the extra particle is more confusing.This is essentially private Api so keep the name `from_reader`.

src/decoder/mod.rs
```diff
@@ -227,6 +227,7 @@ pub struct TiffHeader {
     pub byte_order: ByteOrder,
     pub variant: TiffVariant,
     pub first_ifd: IfdPointer,
+    lenient: bool,
```
197g: This does not make sense to me. The struct is supposed to be public as a description of a file header, the option for the decoding process does not belong here.

src/decoder/mod.rs
```diff
@@ -792,6 +801,11 @@ impl<R: Read + Seek> Decoder<R> {
         Ok(header.open(r))
     }
 
+    pub fn open_with_options(mut r: R, lenient: bool) -> TiffResult<Decoder<R>> {
```
197g: A single `bool` parameter is not what I'd expected from the naming of `_options`. Designing the API to be similar to `Decoder::with_limits`, i.e. configuring the decoder after its creation, would be a much more focused API surface and avoid the need to duplicate a `lenient`-field in all the different ways of creating one. (Also you could choose it on a per-image basis).

.gitignore
```diff
@@ -3,3 +3,4 @@
 **/*.rs.bk
 Cargo.lock
 tests/benches/corpus/
+.DS_Store
```
197g: Please configure this for your own machine (`.git/info/exclude`), not in the PR.

examples/lenient_decode.rs
```diff
@@ -0,0 +1,233 @@
+use tiff::decoder::{BufferLayoutPreference, Decoder, DecodingSampleBuffer, DecodingSampleSlice};
```
197g: I'd be happier with adding parameter parsing to the existing example instead of duplicating it entirely.

197g (commented):
The internal mechanism for this looks better, API could use polish.

soyosaki:
@197g Sorry for the delay. I have fixed the issues you pointed out in your comments. Now one enables lenient mode via `with_lenient` on the decoder, similar to how one currently configures the limits via `with_limits`. The decode example is now modified to take a `--lenient` CLI argument after the file name that will enable lenient parsing mode if the argument exists. Redundant methods and fields are cleared. Please let me know if there's other issues to this PR.

197g (approved):

PR #398, opened 2026-06-27
Add safe Rust zstd decoding backend

Shnatsel:
Use the new `zrip` crate instead of `ruzstd` proposed in  #349. It is over 2x faster and supports multi-frame streams.

The `paranoid` feature is used to disable all `unsafe` code, other than the `fearless_simd` dependency. However I am biased in favor fearless_simd because I worked on it. If you prefer, `fearless_simd` can also be disabled with `default-features=false`, but the amount of `unsafe` in `fearless_simd` is [very low](https://gist.github.com/Shnatsel/61fc294987a1e051ce3835c97dc0fc19) (just safe multiversioning plus a very minimal equivalent of `safe_unaligned_simd`).

The crate is quite young but stands up to fuzzing and roundtrip fuzzing against the C library (i.e. successfully decodes all streams produced by C libzstd under a fuzzer).

The in-tree zstd TIFF decoding test passes with `zstd-safe-rust` feature.

Zstd remains disabled by default. If both `zstd` and `zstd-safe-rust` features are enabled, `zstd` takes precedence.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ fax34 = { package = "fax", version = "0.2.6", optional = true }
 flate2 = { version = "1.0.20", optional = true }
 weezl = { version = "0.1.10", optional = true }
 zstd = { version = "0.13", optional = true }
+ruzstd = { version = "0.8.2", optional = true }
 zune-jpeg = { version = "0.5.1", optional = true }
 image-webp = {version = "0.2.4", optional = true }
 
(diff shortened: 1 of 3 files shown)


lilith:
I'd love a pluggable system for compression and codec libs; I have my own that are pareto in many areas and offer unique safety guarantees in others.

197g (approved):
I'd just be biased for `safe-unaligned-simd` and specific kernels instead but all of that is complementary—if generic SIMD is desirable and works well enough in practice then sure. Gladly the performance difference is not that big of a concern as with `ruzstd`, so yes.

As for a pluggable system I have the same stance as for Simd in the first place, that runtime configuration from a set of enabled options—which may be dynamic—is always nicer than compile time cfg. We need to consider the interface though, the trait we expose for building a decompressor takes inputs and has a currently internal output type (`Box<dyn Read>`) that I found unfortunately restrictive in the past.

Shnatsel:
zrip used to have handwrritten intrinsics for every platform, but then the author deleted them because generic SIMD was just as good and far less of a maintenance burden.

`fearless_simd` also provides safe loads/stores and, unlike `safe_unaligned_simd`, lets you safely call `#[target_feature]` functions. So if explicit intrinsics prove beneficial, they will be very easy to add.

PR #399, opened 2026-07-01
decoder: reconstruct extended-depth samples modulo declared bit depth (supersedes #396)

lilith:
Follow-up to #396, addressing @197g's review point — *"wonder if the library should also clamp the channel values in the non-inverted color case."* This fixes the root cause for both paths, so I'd close #396 in favor of it.

## Problem

Extended grayscale depths promoted to a wider storage type (`Gray(9..=15)` → `u16`, `Gray(17..=31)` → `u32`) can decode to a sample above the declared max `(1<<n)-1`: the horizontal predictor reconstructs `prev + delta` in the promoted storage width (not modulo `2^BitsPerSample`).

In the `WhiteIsZero` path this underflowed `max - v` in `invert_colors` (the #396 panic in debug, wrong pixel in release). In the non-inverted path the value is equally out of range — it just doesn't panic; the output sample keeps junk in its high bits.

## Fix

An n-bit predictor delta is physically stored in n bits, i.e. modulo `2^n`, so reconstruction is defined modulo `2^n` as well. Since `2^n` divides the `u16`/`u32` storage modulus, masking each promoted sample with `(1<<n)-1` once after the predictor yields exactly the mod-`2^n` value — for every sample, including legitimate wrap-arounds — and also bounds any out-of-range value from malformed input. `invert_colors` reverts to a plain subtract (`max - v` is now always in range), superseding the targeted `saturating_sub` from #396.

`mask_samples_to_bit_depth` runs in **every** integer decode branch of `expand_chunk`, not just the extended-depth one. Outside the unpacking branch it is a statically provable no-op — promoted layouts (`declared < storage`) occur only there, because sub-byte depths keep byte-granular storage and reject the horizontal predictor, standard depths have `declared == storage`, and the float predictor requires IEEEFP samples that are never promoted — so the branches cannot diverge if the routing ever changes.

## Note on `min` vs `&`

An earlier revision of this PR clamped with `.min(max)`. That is wrong: saturation mis-decodes legitimate wrap-around deltas. A 12-bit row `[4095, 0, 1]` is stored as deltas `[4095, 1, 1]`; masking reconstructs `[4095, 0, 1]`, but `.min` produced `[4095, 4095, 4095]`. The mask is the only operation that round-trips.

## Tests

Adds exact-pixel regression tests on hand-built TIFFs — 12/20-bit wrap-around (plain and `WhiteIsZero`) and a 9–31-bit encode→decode identity sweep — that fail under saturation and pass under the mask, locking down the modular semantics. Also keeps the two fuzzer-minimized panic seeds under `tests/fuzz_images/oor_panic/` (exercised by the existing `oor_panic` test). Full decode/predict/encode suite, `cargo fmt`, and `clippy` unchanged.


src/decoder/image.rs
```diff
@@ -1198,6 +1198,12 @@ impl Image {
                     scratch,
                 );
 
+                // A promoted extended-depth sample can carry bits above the
+                // declared depth after the predictor (reconstructed in the u16/u32
+                // storage width). Clamp to the declared max so *both* the inverted
+                // and non-inverted paths see in-range values.
+                super::clamp_samples_to_bit_depth(out_row, tiff_bps, out_bps, self.sample_format);
```
197g: The other branches can never hit the issue? Seems about right, `is_all_bits` should not be set and floating point predictor is also an impossible combination. In any case, I'd appreciate a comment or validation to avoid them diverging.

lilith: Right — promoted layouts (declared < storage width) can only arise on the `needs_unpacking` path: sub-byte depths keep byte-granular storage and reject the horizontal predictor, standard depths have declared == storage, and the float predictor requires IEEEFP samples, which are never promoted. So the other branches can't see out-of-range samples today.

Rather than record that only as a comment, I made the tail uniform: the helper is now called in every integer branch — statically a no-op outside the unpacking branch (the `declared < storage` guards can't fire there), so the branches can't diverge if the routing ever changes; the float branch got a comment instead of a dead call.

While making that change I realized `min` was the wrong operation. An n-bit delta is stored in exactly n bits, i.e. modulo 2^n, so reconstruction must reduce modulo 2^n — and since 2^n divides the u16/u32 storage modulus, one mask after the predictor is exact. Saturating instead mis-decodes legitimate wrap-around deltas: 12-bit `[4095, 0, 1]` is stored as deltas `[4095, 1, 1]`, which the clamp decoded as `[4095, 4095, 4095]`. Switched to `& ((1 << n) - 1)`, renamed the helper `mask_samples_to_bit_depth`, and added exact-pixel wrap-around tests (plain and WhiteIsZero) plus a 9–31-bit encode→decode identity sweep on hand-built TIFFs.

Rebased onto current main and retitled the PR to match the new approach.

197g (approved):

Issue #400, opened 2026-07-08
Tiled images with width not a multiple of tile width decode with black left columns


jacovdbergh:
**Version:** tiff 0.11.3 (also reproduces via image 0.25.10's TIFF decoder)

`Decoder::read_image()` corrupts tiled TIFFs whose image width is not a multiple of the tile width. The leftmost tile_width - (width % tile_width) columns decode as zeros on every row except the first row of each tile band (rows 0, 256, 512, … are correct). Reading the same file tile-by-tile with read_chunk() produces correct output, so the bug appears to be in the whole-image assembly, not chunk decoding.

Observed with an uncompressed, pixel-interleaved, 8-bit RGB GeoTIFF, 4000 px wide with 256×256 tiles (4000 % 256 = 160 → leftmost 96 columns black). Widths of 3840 and 4096 (exact multiples) decode correctly.

To reproduce with any input image:

```
# corrupts: 4000 % 256 != 0
gdal_translate -outsize 4000 600 -co TILED=YES -co BLOCKXSIZE=256 -co BLOCKYSIZE=256 in.png bad.tif
# decodes fine: 3840 = 15 * 256
gdal_translate -outsize 3840 600 -co TILED=YES -co BLOCKXSIZE=256 -co BLOCKYSIZE=256 in.png good.tif

let mut d = tiff::decoder::Decoder::new(std::io::Cursor::new(std::fs::read("bad.tif")?))?;
let tiff::decoder::DecodingResult::U8(buf) = d.read_image()? else { panic!() };
let (w, _) = d.dimensions()?;
// pixel (10, 300): expected image content, actual (0, 0, 0)
println!("{:?}", &buf[(300 * w as usize + 10) * 3..][..3]);
```

Expected: pixel values matching the source (GDAL/QGIS/libtiff read the file correctly). Actual: [0, 0, 0] for all x < 96 on every row except each tile band's first row.

We process ortho photo tiffs into tiles for WMTS, and with this bug the result look like this:

<img width="695" height="954" alt="Image" src="https://github.com/user-attachments/assets/1eb7baec-d496-472a-9a1c-69d1067d14e8" />

jacovdbergh:
While debugging this I realised the corruption only occurs when the decoder discards extra samples — our supplier delivers 4-band RGBI tiffs of which we only use the RGB bands. Plain 3-band tiled files with the same non-multiple width decode correctly.

PR #402, opened 2026-07-16
Tolerate missing LZW end-of-information code (fixes #395)

### LINKED ISSUE #395: Tiled LZW TIFFs fail to decode via read_image()
Hello, thank you for maintaining this crate! For the past few years I've been using it to load tif files related to hydrology (DEMs and d8 pointer files), and I've discovered an odd issue when updating from tiff=0.10.0 & weezl=0.1.8 to more recent releases. 

The file I'm trying to load is available online from the United States here (zip 2.6GB, relevant file is ~60MB; there's no need for you to download this file): 
https://prd-tnm.s3.amazonaws.com/StagedProducts/Hydrography/NHDPlusHR/VPU/Current/Raster/NHDPLUS_H_0108_HU4_20220324_RASTER.zip / fdr.tif

This file is compressed and tiled. here's the gdalinfo:
```
$ gdalinfo  ./fdr.tif
Driver: GTiff/GeoTIFF
Files: ./fdr.tif
Size is 13745, 46878
Coordinate System is:
PROJCRS["USA_Contiguous_Albers_Equal_Area_Conic_USGS_version",
    BASEGEOGCRS["NAD83",
        DATUM["North American Datum 1983",
            ELLIPSOID["GRS 1980",6378137,298.257222101,
                LENGTHUNIT["metre",1]],
            ID["EPSG",6269]],
        PRIMEM["Greenwich",0,
            ANGLEUNIT["Degree",0.0174532925199433]]],
    CONVERSION["unnamed",
        METHOD["Albers Equal Area",
            ID["EPSG",9822]],
        PARAMETER["Latitude of false origin",23,
            ANGLEUNIT["Degree",0.0174532925199433],
            ID["EPSG",8821]],
        PARAMETER["Longitude of false origin",-96,
            ANGLEUNIT["Degree",0.0174532925199433],
            ID["EPSG",8822]],
        PARAMETER["Latitude of 1st standard parallel",29.5,
            ANGLEUNIT["Degree",0.0174532925199433],
            ID["EPSG",8823]],
        PARAMETER["Latitude of 2nd standard parallel",45.5,
            ANGLEUNIT["Degree",0.0174532925199433],
            ID["EPSG",8824]],
        PARAMETER["Easting at false origin",0,
            LENGTHUNIT["metre",1],
            ID["EPSG",8826]],
        PARAMETER["Northing at false origin",0,
            LENGTHUNIT["metre",1],
            ID["EPSG",8827]]],
    CS[Cartesian,2],
        AXIS["(E)",east,
            ORDER[1],
            LENGTHUNIT["metre",1,
                ID["EPSG",9001]]],
        AXIS["(N)",north,
            ORDER[2],
            LENGTHUNIT["metre",1,
                ID["EPSG",9001]]]]
Data axis to CRS axis mapping: 1,2
Origin = (1819725.000000000000000,2735005.000000000000000)
Pixel Size = (10.000000000000000,-10.000000000000000)
Metadata:
  DataType=Generic
  AREA_OR_POINT=Area
Image Structure Metadata:
  COMPRESSION=LZW
  INTERLEAVE=BAND
Corner Coordinates:
Upper Left  ( 1819725.000, 2735005.000) ( 72d27'17.90"W, 45d33'49.83"N)
Lower Left  ( 1819725.000, 2266225.000) ( 73d50'31.79"W, 41d29' 0.51"N)
Upper Right ( 1957175.000, 2735005.000) ( 70d45'22.77"W, 45d14'59.72"N)
Lower Right ( 1957175.000, 2266225.000) ( 72d14' 6.51"W, 41d11'22.81"N)
Center      ( 1888450.000, 2500615.000) ( 72d20'29.60"W, 43d22'21.87"N)
Band 1 Block=128x128 Type=Byte, ColorInterp=Gray
  NoData Value=255
```

If i open this file with 0.11.3 with read_image() the library throws: Err("no lzw end code found").

I dug in a bit and tried to make a minimal reprod tif file, it looks like a weird combination of compressed encoded bytes and buffer size. 

It seems that as the reader processes the file row by row it might encounter tiles whose LZW encoded bytes are just the right size to fill the buffer but leave the end code unprocessed. This causes weezl to yield with `LzwStatus::NoProgress` even though `consumed_out > 0`. This should have been a successful partial read, but it currently throws.

I think this got introduced a while back (like a year?) when the LZW config started including ` .with_yield_on_full_buffer(true);` I didn't catch it till now because i've been pinned on 0.10.

Any output is a successful partial read from weezl, so we could fix this issue in a way that doesn't depend on the with_yield_on_full_buffer state at all by just returning ok if we have output.

```diff
#[cfg(feature = "lzw")]
impl<R: Read> Read for LZWReader<R> {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        loop {
            let result = self.decoder.decode_bytes(self.reader.fill_buf()?, buf);
            self.reader.consume(result.consumed_in);

+            if result.consumed_out > 0 {
+                return Ok(result.consumed_out);
+            }

            match result.status {
            ...
            }
      }
}
...
```

I'd be glad to open a PR if it helps but i thought we could discuss here first.

I've attached a minimal tif file with the tiling and encoded bytes that throw the error in case you don't want to download a 2GB mostly irrelevant zip from the usgs. Either way, here's a minimal runnable example snippet to run either the fdr.tif file from the usgs or my min_repro.tif (1Kb, attached) and see the error:

```rs
use std::env;
use std::fs::File;
use tiff::decoder::{Decoder, Limits};

fn main() {
    let path = env::args().nth(1).expect("usage: trigger_lzw_bug <path.tif>");
    let file = File::open(&path).unwrap();
    let mut dec = Decoder::new(file).unwrap().with_limits(Limits::unlimited());
    dec.read_image().unwrap();
}
```

Currently the fix that is working for me is to pin tiff = "=0.10.0"

[min_repro.tif](https://github.com/user-attachments/files/28649573/min_repro.tif)

kleinicke:
## Problem

`Decoder::read_image()` (and any code path that reads a strip/tile through
`LZWReader`) fails with `IoError(UnexpectedEof, "no lzw end code found")` when
an LZW-compressed strip or tile's compressed byte stream runs out without a
terminal end-of-information (EOI) code, even though every real pixel byte
that strip/tile is supposed to contain was already present and decoded
correctly before the input ran out.

This is issue #395: "Tiled LZW TIFFs fail to decode via `read_image()`".
libtiff-based tools (including `tifffile`) decode the same files without
complaint.

## Root cause — two layers

**What broke for the files in #395 (0.10.0 → 0.11.x regression):** with
`weezl` 0.1 (as pinned by every released `0.11.x`), `decode_bytes` configured
with `.with_yield_on_full_buffer(true)` can report `LzwStatus::NoProgress`
when the caller's output buffer fills at a point where the EOI code is still
unprocessed in the input — exactly the situation on tiles whose decoded size
is an exact multiple of the read granularity. `LZWReader::read` treats *any*
`NoProgress` as a hard error, so perfectly valid, EOI-terminated tiles failed
with "no lzw end code found". This matches the reporter's diagnosis in #395.

On current `master` this specific failure no longer reproduces: commit
`0fdf1e0` ("Upgrade weezl to 0.2") changed the status semantics so that a
full output buffer yields `Ok` instead of `NoProgress`. We verified the
bisection empirically: at `0fdf1e0^` the files from our regression corpus
fail with the exact #395 error, and swapping *only* the `weezl` dependency to
0.2.1 at that same commit makes them decode pixel-perfectly. So #395's
headline symptom was fixed incidentally, without a test pinning it.

**What is still broken on `master`:** the `NoProgress` arm still hard-errors
whenever the compressed input is exhausted before the decoder has seen an EOI
code *and* the caller asks for more bytes. That is the correct outcome for a
genuinely truncated stream — but it also fires for streams that contain every
meaningful byte and merely omit the trailing EOI (and, on padded edge tiles,
the final padding bytes), which real-world encoders are known to do on the
last coding unit of an image. libtiff accepts such streams: it decodes until
the strip/tile's known-in-advance output size is satisfied and never insists
on seeing an EOI code.

## The fix

`src/decoder/stream.rs`, `LZWReader::read`: when `NoProgress` occurs, check
whether the underlying (length-limited) reader is actually exhausted. If it
is, treat this the same way as the normal end of a well-formed stream and
return `Ok(0)` (EOF) instead of an error. If there is still compressed input
available and the decoder nonetheless can't make progress, that's a
different, still-real bug, so the original hard error is preserved for that
case.

```rust
Ok(weezl::LzwStatus::NoProgress) => {
    if self.reader.fill_buf()?.is_empty() {
        return Ok(0);
    }
    return Err(io::Error::new(
        io::ErrorKind::UnexpectedEof,
        "no lzw end code found",
    ));
}
```

This also removes the reader's dependence on the *presence* of an EOI code
altogether, so a future change in `weezl`'s status semantics can no longer
reintroduce the 0.11.x regression.

### The error-vs-truncation ambiguity is intentional and libtiff-compatible

A stream that is missing only its EOI code, and a stream that is genuinely
truncated in a way that happens to run out of input at exactly a
buffer-satisfying boundary, are **not** distinguishable from inside the LZW
layer. This patch does not try to distinguish them locally — it can't, and
neither can libtiff. Instead it defers entirely to the caller's own
exact-size read, driven by the strip/tile geometry that is known
independently of the compressed stream: if the caller already has every byte
it asked for when input runs out, decoding is complete; if it still needs
more bytes, `read_exact` surfaces `UnexpectedEof` on its own, so a genuinely
short/corrupt stream is still rejected. This defer-to-known-output-size
strategy is exactly what libtiff's LZW decoder does.

## The tests

`tests/lzw_missing_eoi.rs` — all fully synthetic, built in memory; no binary
test images are vendored. The missing-EOI streams are produced by LZW-encoding
a pixel buffer with `weezl` and trimming the encoded stream to the shortest
prefix that still fully decodes back to the original pixels (self-verifying:
the trim provably removes only the EOI code and its byte-alignment padding,
never real data).

- `tiled_lzw_missing_final_padding_and_eoi_still_decodes` — **fails verbatim
  with "no lzw end code found" before this fix, passes after.** A 20×32 image
  in a single 32×32 tile: every tile row carries 12 trailing padding bytes
  the decoder must read-and-discard, and the crafted stream omits both the
  final row's padding and the EOI code — the encoder-shortcut shape of #395.
- `lzw_strip_missing_eoi_code_still_decodes` — a single-strip image whose
  stream is missing only the EOI code. On current `master` (with weezl 0.2)
  this already decodes; the test pins that behavior so the 0.10.0 → 0.11.x
  regression can't silently come back.
- `lzw_strip_truncated_mid_stream_still_errors` — the (already EOI-less)
  stream additionally truncated by 25%, so real pixel data is missing.
  Asserts decoding still errors; guards against the fix becoming "ignore any
  decode shortfall".

`cargo test` (all integration suites and doctests) is green; `cargo fmt --
--check` is clean; `cargo clippy --all-features --all-targets -- -D warnings`
reports only one pre-existing, unrelated `#[expect(clippy::unnecessary_unwrap)]`
lint-version mismatch in `src/decoder/mod.rs:849`, confirmed present on
unmodified `master` with the same local toolchain.

A `CHANGES.md` entry is included under "Version 0.12.0 (unreleased) / Fixes".

## Real-world verification

Verified against tiled, LZW-compressed sample TIFFs from an external
regression corpus (each with an independently produced ground-truth pixel
fixture): they fail with the exact #395 error on the released `0.11.3` and
decode pixel-perfectly on this branch.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,14 @@ Compatibility:
   from either StripOffsets or TileOffsets with the latter taking precedence.
   This is closer to the behavior of libtiff.
 
+Fixes:
+- LZW strips/tiles whose compressed stream ends without an explicit
+  end-of-information (EOI) code are now tolerated instead of failing with
+  "no lzw end code found", as long as all data implied by the strip/tile
+  geometry has been produced. This matches libtiff, which stops at the
+  expected output size regardless of a trailing EOI code. Genuinely
+  truncated streams still fail to decode.
+
 # Version 0.11.3
 
 Additions:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,10 @@ criterion = "0.3.1"
 crc32fast = "1.5"
 libtest-mimic = "0.8.1"
 walkdir = "2.5"
+# Used directly by tests/lzw_missing_eoi.rs to build a synthetic LZW stream
+# with a deliberately omitted end-of-information code; version kept in sync
+# with the `weezl` entry in `[dependencies]`.
+weezl = "0.2.1"
 
 [features]
 default = ["deflate", "fax", "jpeg", "lzw"]
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -170,6 +170,34 @@ impl<R: Read> Read for LZWReader<R> {
                     }
                 }
                 Ok(weezl::LzwStatus::NoProgress) => {
+                    // `NoProgress` means the decoder consumed all available input and
+                    // produced no new output this round (`consumed_out` is always `0`
+                    // here; see `weezl`'s own `LzwStatus::NoProgress` invariant). This
+                    // happens in two situations that are indistinguishable from here:
+                    //
+                    //   1. The stream is missing its end-of-information (EOI) code.
+                    //      Some encoders omit it on the final strip/tile of an image
+                    //      (see <https://github.com/image-rs/image-tiff/issues/395>
+                    //      for a real-world tiled LZW TIFF exhibiting this). libtiff
+                    //      does not require the EOI code: it decodes until the strip's
+                    //      known-in-advance output size is reached and stops there,
+                    //      regardless of whether an EOI code follows.
+                    //   2. The stream is genuinely truncated or corrupt and simply
+                    //      ends before producing all the data the caller expects.
+                    //
+                    // We cannot tell these apart locally, and deliberately don't try:
+                    // this is the same ambiguity libtiff accepts. Instead, treat
+                    // input exhaustion as a normal end-of-stream (`Ok(0)`) rather than
+                    // a hard error, the same way a well-formed EOI-terminated stream
+                    // would be reported once fully consumed. The caller (e.g.
+                    // `Read::read_exact`, driven by the strip/tile geometry that is
+                    // already known independently of this stream) still surfaces an
+                    // `UnexpectedEof` on its own if fewer bytes were actually produced
+                    // than expected, so case 2 above is still reported as an error --
+                    // just by the caller instead of by this reader.
+                    if self.reader.fill_buf()?.is_empty() {
+                        return Ok(0);
+                    }
                     return Err(io::Error::new(
                         io::ErrorKind::UnexpectedEof,
                         "no lzw end code found",
--- a/tests/lzw_missing_eoi.rs
+++ b/tests/lzw_missing_eoi.rs
@@ -1,0 +1,384 @@
+//! Regression test for <https://github.com/image-rs/image-tiff/issues/395>:
+//! LZW-compressed strips/tiles whose encoder omitted the end-of-information
+//! (EOI) code previously failed to decode with "no lzw end code found",
+//! even though every real pixel byte was present in the stream and the
+//! output size is already known in advance from the strip/tile geometry.
+//! libtiff tolerates this; see the fix in `src/decoder/stream.rs`
+//! (`LZWReader::read`).
+//!
+//! The test file below is built entirely in memory (classic, single-strip,
+//! 8-bit grayscale TIFF) rather than vendored from a real-world sample, so
+//! the LZW payload can be crafted to omit the EOI code deterministically.
+
+extern crate tiff;
+extern crate weezl;
+
+use std::io::Cursor;
+use tiff::decoder::{Decoder, DecodingSampleBuffer};
+
+/// Fully LZW-encodes `pixels` (MSB-first bit packing, TIFF early code-size
+/// switch, including the terminal end-of-information code) and returns the
+/// resulting well-formed, EOI-terminated byte stream.
+fn lzw_encode_with_eoi(pixels: &[u8]) -> Vec<u8> {
+    weezl::encode::Encoder::with_tiff_size_switch(weezl::BitOrder::Msb, 8)
+        .encode(pixels)
+        .expect("encoding raw bytes cannot fail")
+}
+
+/// Attempts to decode `stream` (MSB-first, TIFF early code-size switch) into
+/// exactly `expected.len()` bytes, tolerating a missing EOI code exactly the
+/// way the fixed `LZWReader` does (stop once the known output size is
+/// reached, or once input runs out with nothing left to decode). Returns
+/// `Some(decoded_bytes)` on success, `None` if the stream doesn't decode to
+/// the exact expected length.
+fn try_decode_lzw_tolerant(stream: &[u8], expected_len: usize) -> Option<Vec<u8>> {
+    let mut decoder = weezl::decode::Decoder::with_tiff_size_switch(weezl::BitOrder::Msb, 8);
+    let mut out = vec![0u8; expected_len];
+    let mut in_pos = 0usize;
+    let mut out_pos = 0usize;
+    while out_pos < expected_len {
+        let result = decoder.decode_bytes(&stream[in_pos..], &mut out[out_pos..]);
+        in_pos += result.consumed_in;
+        out_pos += result.consumed_out;
+        match result.status {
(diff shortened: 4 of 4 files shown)


kleinicke:
The CI failures here are pre-existing and unrelated to this diff: all four failing jobs error on the same new clippy 1.97 lint (`useless_borrows_in_formatting`) at `src/decoder/ifd.rs:366`, a file this PR doesn't touch — it fails identically on unmodified `main`. Opened #404 with the one-character fix; once that lands I'll rebase to re-run CI here.

197g:
Duplicate of #391 but much more verbose. And not more accurate? While it's true that `read_exact` surfaces _some_ error it changes the error so that the two causes are now conflated. Also, we're already supposed to be calling `read_exact` with the expected sample buffer size and `weezl = 0.2` never produces NoProgress together with output. As far as I can tell, the premise of the fix is flawed and I've already commented so in the other PR. Rather the problem is likely identified with #400 (and #401?) as a confusion about the number of expected padding. That would run through `io::copy(…)` instead which indeed now suppresses short reads. Case in point, I have a fork where I'm trying to get rid of io traits for decompression that _accurately_ reads the padding and tests still fail on it.

kleinicke:
Agreed that #391 covers the code change, and your padding-accounting reading matches what I measured: on current `main`, a tiled LZW stream that contains every pixel byte but omits the final tile's padding bytes + EOI still fails with `no lzw end code found` — the error fires in the padding-discard step, not while producing samples. The synthetic tests in this branch reproduce that without binary fixtures, in case they're useful for your decompression rework.

For real-world test data: I used https://github.com/tlnagy/exampletiffs and went through every file checking the decoded pixels against libtiff ground truth — the `shapes_lzw_tiled*.tif` files there hit this path, and that sweep is also how #403 (tiled+planar panic) turned up.

PR #407, opened 2026-07-25
feat: decode old-style (pre-TIFF-6.0) LZW streams

aarshivv:
Support old-style (pre-TIFF-6.0) LZW

Compression = 5 actually covers two LZW variants: the current one, and the older pre-6.0 encoding that some old files still use. There's no tag to tell them apart, so right now the decoder assumes new-style and bails out with InvalidCode on the old ones — quad-lzw.tif from the libtiffpic set is a good example (it was sitting in the test list as a skipped decode).

The two differ in bit order and code-size timing: new-style is MSB-first with the "early change" size switch, old-style is LSB-first without it. This detects which one a stream is using the same way libtiff does — old-style begins with the clear code as 0x00 followed by an odd byte, whereas new-style starts with 0x80 — and configures weezl accordingly. If the peek is inconclusive it just falls back to the modern decoder, so nothing changes for regular files.

I pinned the exact variant empirically rather than guessing: decoding every strip of quad-lzw.tif, only LSB + no-size-switch decodes all of them cleanly, and the result is pixel-identical to libtiff. As a nice cross-check, it produces the same hash as quad-tile.tif, which is the tiled version of the same image and already passing.

Tests:
- quad-lzw.tif in decode_libtiffpic goes from skipped to a verified hash.
- Added test_old_style_lzw using the existing (previously unused) quad-lzw-compat.tiff fixture.
- Ran it against every LZW file I could find (libtiffpic + a pile of real-world scans) — old-style ones now decode, and all the new-style files are byte-for-byte unchanged.

### DIFF
--- a/src/decoder/stream.rs
+++ b/src/decoder/stream.rs
@@ -141,14 +141,32 @@ pub struct LZWReader<R: Read> {
 impl<R: Read> LZWReader<R> {
     /// Wraps a reader
     pub fn new(reader: R, compressed_length: usize) -> LZWReader<R> {
-        let configuration =
+        let mut buffered = BufReader::with_capacity(
+            (32 * 1024).min(compressed_length),
+            reader.take(u64::try_from(compressed_length).unwrap()),
+        );
+
+        // `Compression = 5` covers both the current LZW variant and the pre-TIFF-6.0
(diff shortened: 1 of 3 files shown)


197g (approved):
Not a fan of the heuristic but given libtiff uses it seems okay. And yes, newstyle is unique so there's no regression.

PR #409, opened 2026-08-09
feat: LERC decode

micahcc:
This might be debatable because its sort of niche, but it would really help with geotiffs I'm working on

- Spec: https://github.com/esri/lerc
- add support for LERC decoding as produced by gdal `COMPRESS=LERC`
- pretty straightforward, adds a dep though :/


src/decoder/stream.rs
```diff
@@ -528,6 +528,80 @@ impl Read for Group3Reader {
...
+                    crate::TiffUnsupportedError::UnsupportedCompressionMethod(
+                        crate::tags::CompressionMethod::Lerc,
+                    ),
+                ));
+            }
+        };
+
+        let decoded = lerc_reader::decode(&decompressed)
+            .map_err(|e| io::Error::new(io::ErrorKind::InvalidData, e))?;
+
```
197g, line 576: There seems to be no way to limit the number of pixels [and it trusts the info](https://docs.rs/lerc-reader/latest/src/lerc_reader/lerc1.rs.html#194-203) contained in the input data instead of verifying against the chunk size expected from Tiff. Even though there's an internal limit (512MB right now) that's a pretty direct DOS vector through the allocator in calls that expected to run below this sizeable default.

From a glance at the documentation, [`decode_band_set_into`](https://docs.rs/lerc-reader/latest/lerc_reader/fn.decode_band_set_into.html) avoids both. Alternatively checking the `BlobInfo` ahead of time should also work as `dimensions` are available for the constructor, too. That would also allow larger images to be decoded as well, I think?

Note: the main allocation bounded by `compressed_bytes` is not affected since `image.rs` checks it against `intermediate_buffer_size` already.

197g (commented):
Pretty sure the extra optional dependencies aren't an issue per-se. That said, the specific crate is quite new and some parts / commits look rather vibe-coded so there are some concerns in terms of stability / eventual replacement. Doesn't have to be though and I could be wrong—the interface is not bad and we consume a rather small part of it.

src/decoder/stream.rs
micahcc: oof, good callout on limiting pixel count. Will look into it. 

Yeah agree on the vibe-coded concerns, the same dev also has a full geotiff crate, which in some ways may fit my needs better, but the only way to have created it this quickly is through LLM usage. I don't have anything particularly against it, but I'm also guilty of writing a lot of code that meets my need and never touching it again.

My first look at LERC is that it is mostly just a quantization heuristic, but claude seemed to think it would take a lot of lines of code to inline. I can give it a shot just to see.

micahcc: Yeah so seems like it would end up being 2000ish lines of code to inline the reading portion. Thoughts on which would be better? Choices are basically adding a `src/decoder/lerc_reader/*.rs` that just ports the reading versus using the third party crate? 

If we wanted to pass in the higher level TIFF shape into the LERC reader that would require PRs into lerc-reader . I suppose the real test of that crates stability is if they take PRs lol but its a pain in the butt to push a PR there to get this PR into here.

micahcc:
As noted in the issue (https://github.com/image-rs/image-tiff/issues/410#issuecomment-5317358177 ): LERC seems to give worse results than a simpler alternative. Might need to add geotiff scaling support, but that would be a separate (simpler) PR. Given the complexity involved I don't think its worth the maintenance burden unless someone finds a really crucial use case.

PR #412, opened 2026-08-25
Fix panic decoding tiled TIFFs with PlanarConfiguration=Separate

### LINKED ISSUE #403: Decoding a tiled + PlanarConfiguration=2 TIFF panics in Image::expand_chunk (assertion at src/decoder/image.rs:1153)
## Affected version

- Crate `tiff`, master @ `8ad05b85dbc488caa17e0b75b35033b81ca24ff5` (merge of #397, 2026-07-03). Also present on the `0.11.3`-adjacent history — the buggy code (`TileAttributes::get_padding`) is unrelated to #397 and predates it.
- Reproduces with both `Compression::None` and `Compression::Lzw`; compression is not the trigger.

## Summary

Any TIFF that is **tiled** (`TileWidth`/`TileLength` present) *and* uses **`PlanarConfiguration = 2`** (separate planes) panics while decoding, as soon as the image has more than one tile row or column in a plane (i.e. an "edge" tile is not the very first tile of the whole file). Striped planar images decode fine — only the tiled+planar combination is affected.

Panic:
```
thread 'main' panicked at src/decoder/image.rs:1153:9:
assertion failed: buf.len() >= layout.row_stride * (data_dims.1 as usize - 1) + data_row_bytes
```

## Root cause

`TileAttributes::get_padding` (`src/decoder/image.rs:45-61`) computes a tile's row/column purely from its *raw* tile index:

```rust
pub fn get_padding(&self, tile: usize) -> (usize, usize) {
    let row = tile / self.tiles_across();
    let column = tile % self.tiles_across();
    ...
}
```

`tiles_across()`/`tiles_down()` (lines 33-38) describe the grid of a **single plane**. But the caller, `chunk_data_dimensions` (`src/decoder/image.rs:1003-1011`, specifically the call at line 1005), is invoked with a **global** chunk index that already has the per-plane offset baked in:

```rust
// src/decoder/mod.rs, Decoder::read_image_bytes (~line 1857-1866)
for chunk in 0..readout.chunks_per_plane {
    ...
    for (idx, &plane_offset) in plane_offsets[..used_plane_offsets].iter().enumerate() {
        let chunk = chunk + idx as u32 * readout.chunks_per_plane; // <-- global index across planes
        ...
        image.expand_chunk(value_reader, ..., readout, chunk, scratch)?;
    }
}
```

For plane `idx >= 1`, `chunk` is offset by `idx * chunks_per_plane`, so `get_padding` divides/mods a value outside the single-plane grid it was designed for. E.g. for a 128x72 image with 32x32 tiles (4x3 tiles/plane = 12 tiles/plane) and chunk 32 (plane 2, tile row 2 = last row within that plane):

```
row    = 32 / tiles_across() = 32 / 4 = 8   // should be 32 % 12 / 4 = 2
column = 32 % tiles_across() = 32 % 4 = 0   // should be (32 % 12) % 4 = 0 (ok by luck)
```

`row == 8` never equals `tiles_down() - 1 == 2`, so `padding_down` is silently computed as `0` — the bottom-edge tile is not recognized as such, and `chunk_data_dimensions` returns the full, unclamped `tile_length` (e.g. 32) instead of the true remaining row count (e.g. 8).

`expand_chunk` (`src/decoder/image.rs:1153`) then asserts the output buffer is large enough for that (wrongly inflated) `data_dims.1`, and since the real remaining space in the output buffer was correctly sized for the *true* (smaller) edge-tile height, the assertion fails.

`get_padding` needs to operate on the chunk index **modulo `chunks_per_plane`** (or equivalently the caller should pass a per-plane-relative index), not the raw global index that includes the plane offset.

Because `PlanarConfiguration::Chunky` never offsets the chunk index by plane (`chunks_per_plane == total tile count`, loop always uses `idx == 0`), chunky tiled images are unaffected — this is specific to the planar + tiled combination. Striped planar is also unaffected because `chunk_data_dimensions`'s `ChunkType::Strip` arm (lines 987-1002) derives padding via `chunk_index % strips_per_band`, which already reduces modulo the per-plane strip count.

## Minimal synthetic reproducer (no attachments needed)

Generate a tiny fully-synthetic tiled+planar TIFF with Python/tifffile (any tile size smaller than the image in both dimensions triggers it, since it just needs one edge tile in plane > 0):

```python
import numpy as np
import tifffile

# 24x24 RGB, 16x16 tiles -> 2x2 tile grid per plane (edge tiles on
# right column and bottom row), PlanarConfiguration=2 (separate planes).
data = np.zeros((3, 24, 24), dtype=np.uint8)
for c in range(3):
    data[c] = (np.arange(24 * 24).reshape(24, 24) + c * 7) % 256

tifffile.imwrite(
    "tiled_planar_repro.tif",
    data,
    tile=(16, 16),
    planarconfig="separate",
    photometric="rgb",
    compression=None,
)
```

Decode with upstream's own example:

```rust
// examples/decode.rs (unmodified)
let mut reader = tiff::decoder::Decoder::open(std::io::BufReader::new(
    std::fs::File::open("tiled_planar_repro.tif")?,
))?;
reader.next_directory()?;
let mut data = tiff::decoder::DecodingSampleBuffer::I8(vec![]);
reader.read_image_to_buffer(&mut data)?; // panics
```

```
$ cargo run --example decode -- tiled_planar_repro.tif
thread 'main' panicked at src/decoder/image.rs:1153:9:
assertion failed: buf.len() >= layout.row_stride * (data_dims.1 as usize - 1) + data_row_bytes
stack backtrace:
   ...
   3: tiff::decoder::image::Image::expand_chunk
             at ./src/decoder/image.rs:1153:9
   4: tiff::decoder::Decoder<R>::read_image_bytes
             at ./src/decoder/mod.rs:1868:23
   5: tiff::decoder::Decoder<R>::read_image_to_buffer
             at ./src/decoder/mod.rs:1824:14
```

Also independently reproduced with two real-world 128x72 RGB / 32x32-tile files (one LZW-compressed, one uncompressed) — same panic, same line, confirming compression is irrelevant.

## Notes for the fix

- `TileAttributes::get_padding` should take (or the caller should compute) a chunk index relative to the current plane, e.g. `chunk_index % chunks_per_plane`, before doing `/ tiles_across()` and `% tiles_across()`.
- Given the existing `// TODO: Should these return errors instead?` comment directly above the assertions at `src/decoder/image.rs:1151-1153`, it may also be worth turning these into proper `TiffError`s rather than panicking, as a defense-in-depth measure for any other buffer-sizing miscalculation.

youdie006:
Fixes #403.

Decoding a tiled TIFF with `PlanarConfiguration=Separate` panics whenever a plane's tile grid is more than one tile in either direction.

`TileAttributes::get_padding` (`src/decoder/image.rs:45`) derives the tile's row and column from the **single-plane** grid:

```rust
let row = tile / self.tiles_across();
let column = tile % self.tiles_across();
```

but the callers pass a **global** chunk index that already includes the per-plane offset (`chunk + idx * chunks_per_plane`). For any plane past the first, that index falls outside the single-plane grid, so a bottom-edge tile is not recognised as one, `padding_down` comes back as 0, `chunk_data_dimensions` reports an over-tall tile, and `expand_chunk` trips its buffer assertion:

```
thread 'test_tiled_planar_rgb_u8' panicked at src/decoder/image.rs:1169:9:
assertion failed: buf.len() >= layout.row_stride * (data_dims.1 as usize - 1) + data_row_bytes
```

## Oracle

The strip path already does this reduction — it indexes with `chunk_index % strips_per_band` — while the tile path passes the raw global index through. Same class of index, two different treatments in the same decoder.

## Fix

One line, reducing the global index to the plane-relative grid before the row/column math. The product is at least 1, so there is no division by zero, and it is a no-op for chunky and single-plane images, where the global index already equals the plane-relative one.

## Test

`tests/images/tiled-planar-rgb-u8.tif` is a new 24x24 RGB u8 fixture: 16x16 tiles, so a 2x2 grid per plane, `PlanarConfiguration=Separate`, uncompressed. `test_tiled_planar_rgb_u8` decodes it through the multi-plane path.

Worth noting for review: the plain `read_image()` path only reads plane 0 and therefore does **not** trigger this, so the test goes through `read_image_to_buffer`.

Verified red then green — with the one-line fix removed and the test kept, it panics as shown above; with the fix it passes. Full suite green: 23 + 1 + 5 + 1 + 77 + 61 passing, 0 failed. `cargo fmt -- --check` clean and `cargo clippy --no-deps --all-features --all-targets -- -D warnings` clean, matching the two CI invocations.

Credit to @kleinicke for the report.

Disclosure: I use AI assistance in my work, and I review and verify everything before it goes out. The red-green runs above are my own.


youdie006:
Flagging that the red CI here is not from this change, so it does not need to hold up review.

`clippy` and `features (default,zstd)` both fail on `using chunks_exact with a constant chunk size` (12 occurrences, plus 4 of `chunks_exact_mut`). That is a lint that became deny-by-default in the rustc 1.98.0 toolchain the workflow now pulls, firing on pre-existing code — this diff does not add, move, or touch any `chunks_exact` call site.

The same failure is on every open PR whose CI has run since the toolchain moved (#411, #391, #389 alongside this one); the two that show green, #388 and #330, last ran before it. The remaining red jobs in the matrix are `CANCELLED`, not failures — `fail-fast` cancelling the siblings once `default,zstd` failed.

Happy to send the `chunks_exact` cleanup as a separate PR if that would help unblock the queue — it seemed wrong to fold an unrelated repo-wide lint sweep into a decoding fix.

youdie006:
The red CI here is not from this PR.

`clippy` fails on `clippy::chunks_exact_to_as_chunks`, which is new in Rust 1.98 (the lint page
the log links to is `rust-1.98.0/index.html#chunks_exact_to_as_chunks`). It fires on
`src/decoder/ifd.rs:602,612,622,632,642,652,660,671,684,694,704,714`, `src/decoder/mod.rs:297`
and `src/tags.rs:626,632,638` — all pre-existing `.chunks_exact(2)` / `.chunks_exact(4)` calls.
This PR does not touch any of those files, and `src/decoder/image.rs` never appears in the log.

Running the CI command locally at both ends:

```
$ cargo clippy --no-deps --all-features --all-targets -- -D warnings
origin/main   exit 0, no warnings
this PR       exit 0, no warnings
```

(on 1.97.0, which predates the lint — which is also why it passes at both ends.)

The `features (default,zstd)` job runs the same clippy step, so it fails for the same reason,
and the other `features` jobs were cancelled by the matrix rather than failing on their own.

Happy to rebase once the lint is addressed on main, or to fold the `as_chunks` migration into
this PR if you would rather have it here — though that would touch a lot of unrelated lines, so
I have left it out.

fintelia:
That's an excessive amount of text/explanation for a 1 line fix. Feel free to open a new PR with these changes after reviewing [our LLM policy](https://github.com/image-rs/organization/blob/main/CONTRIBUTING.md#llm-policy)

